PR #7, opened 2023-11-15
Initial plug-in with support for single and ensemble runs of NESO solvers

### LINKED ISSUE #1: Create task for running Electrostatic2DV solver with two stream test configuration
As an initial test case, we should [create a FabSim3 task](https://fabsim3.readthedocs.io/en/latest/create_new_plugin/#write-application-specific-functionalities) which allows execution of the NESO [`Electrostatic2D3V` solver](https://github.com/ExCALIBUR-NEPTUNE/NESO/tree/main/solvers/Electrostatic2D3V) with [the two stream test configuration](https://github.com/ExCALIBUR-NEPTUNE/NESO/tree/main/test/integration/solvers/Electrostatic2D3V/TwoStream) also used as [the example configuraiton in the `NESO-UQ` repository](https://github.com/ExCALIBUR-NEPTUNE/NESO-UQ).

Ideally this should expose the same set of parameters as used in the NESO-UQ example (and using the same defaults) as optional arguments to the task that is:

- `num_particles_total`: number of particles (~Monte Carlo samples) (integer, default 400000)
- `particle_initial_velocity`: initial particle velocity (float, default 1.0)
- `particle_charge_density`: particle charge density (float, default  105)
- `particle_number_density`: particle number density (float, default 105)
- `particle_position_seed`: seed fpr pseudo-random number generator used to set initial particle states (integer, default 284289374)

with these values being written to a templated conditions file [similar to this](https://github.com/ExCALIBUR-NEPTUNE/NESO-UQ/blob/main/examples/two_stream/two_stream.template) so that to for example run with 10000 particles and initial particle velocity 1.1 we would run something like

```
fabsim localhost run_electrostatic2d3v_two_stream:num_particles_total=10000,particle_initial_velocity=1.1
```

The [FabNEPTUNE plug-in](https://github.com/UCL-CCS/FabNEPTUNE) might be a useful reference.

### LINKED ISSUE #2: Add task for running simulations with `H3LAPD` solver
From discussions with NESO team, seems like H3LAPD should be a more computationally challenging example with both relatively complex continuum / Nektar++ and particle-in-cell components.

- [ ] Get H3LAPD solver running locally with example configurations.
- [ ] Figure out which parameters from conditions file to expose (both 'model' parameters affecting simulation behaviour and 'control' variables affecting simulation performance / accuracy)

### LINKED ISSUE #3: Add task for running ensembles of `Electrostatic2D3V` solver

dleggat:
Implemented a FabNESO plugin that supports single and ensemble runs for NESO solvers.
The choice of solver and conditions files are abstracted at point of use.
Utility files are provided to create new directories - including for ensemble running - and encoding the conditions files to vary the model input parameters.


matt-graham:
Thanks for opening this!

For the linting, if you install `pre-commit` in your local development environment with

```
pip install pre-commit
```
or `conda install -c conda-forge pre-commit` then you can run

```
pre-commit install
```

from the root of the repository to set-up `pre-commit` locally with the configuration in [`pre-commit-config.yaml`](https://github.com/UCL/fabneso/blob/main/.pre-commit-config.yaml). That will then run the various hooks defined there every time you commit which should ensure the linting is done automatically. In particular it looks like the checks are currently failing because the code is getting changed by `black`.

Also as a more general comment: typically for [Python projects we try to adhere to PEP8 code style](https://github-pages.arc.ucl.ac.uk/coding-standards/python/), in particular using `snake_case` for variable and function names and `CamelCase` only for class names. Module names should also generally be in `snake_case` too.

FabNeso.py
```diff
@@ -0,0 +1,58 @@
...
+# Please refer to LICENSE for detailed information regarding the licensing.
+#
+# This file contains FabSim definitions specific to FabNeso
+#
+# authors: Duncan Leggat
+
+try:
+    from fabsim.base.fab import *
+except ImportError:
+    from base.fab import *
```
matt-graham: Realise this is just taken from `FabDummy` example, but I would avoid `from ... import *` (and in fact Ruff may well disallow this) imports in favour of just importing the specific names you need, so for example something like

```Python
from fabsim.base.fab import (
    add_local_paths,
    execute,
    find_config_file_path,
    job,
    run_ensemble,
    task,
    update_environment,
    with_config, 
)
```
or just use import the relevant module and use qualified names e.g.

```Python
try:
    import fabsim.base.fab as fab
except ImportError:
    import base.fab as fab
```
then use `fab.add_local_paths`, `@fab.task` etc.

FabNeso.py
```diff
@@ -0,0 +1,58 @@
...
+    execute(put_configs, config)
+
+    env.neso_solver = solver
+
+    #This we presumably change somehow so that it gets changed throughout
+    #the SWEEP dir?
+    env.neso_conditions_file = (find_config_file_path(config)+
+                                "/" + conditionsFileName)
+    #All of these should be in a config file somewhere
+    env.neso_mesh_file = find_config_file_path(config) + "/" + meshFileName
```
matt-graham: Ideally when constructing paths, we should either use `pathlib.Path` objects which overload `/` operator to join path components, or use `os.path.join`, as this make sure the paths are robust to operating system differences in path separators etc.

So for example if we add a `from pathlib import Path` in the import section, here would do something like

```Python
    env.neso.mesh_file = Path(find_config_file_path(config)) / meshFileName
```

FabNeso.py
dleggat: When I do this, I get new errors from mypy complaining when I redefine the same paths as: 
```python
except ImportError:
    from base.fab import (
        add_local_paths,
    )
``` 
with an error:
```
FabNeso.py:25: error: Name "add_local_paths" already defined (possibly by an import)  [no-redef]
``` 
Did this come up for you if you define them in both the try and except stages?

matt-graham: I ended up using the latter 

```Python
try:
    import fabsim.base.fab as fab
except ImportError:
    import base.fab as fab
```

approach then qualifying names with `fab.` to avoid the duplication of names being imported.

dleggat: In the end the issues seemed to come because I'd initially set up the repository in the plugin directory within FabSIM itself - I moved the plugin to another directory elsewhere on my computer and ran it again and the import errors were no longer raised. This also solved the errors I was getting from mypy in the fabsim base, which it no longer raises.

FabNeso.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
```
matt-graham:
```suggestion
```
Realise this is from FabDummy example again, but these encoding comments are only needed in Python 2.x and we'll only be supporting Python 3.x so can be safely removed.

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+#
+# This source file is part of the FabSim software toolkit, which is
+# distributed under the BSD 3-Clause license.
+# Please refer to LICENSE for detailed information regarding the licensing.
+#
+# This file contains FabSim definitions specific to FabNeso
+#
+# authors: Duncan Leggat
```
matt-graham:
```suggestion
"""
Task definitions for FabNESO plug-in to FabSIM software toolkit.

Defines tasks for running simulations using Neptune Exploratory Software (NESO).
"""
```

Typical convention is to document what a module does using a multiline string (`""" ... """`) which will get assigned to `__doc__` attribute (and used in autogenerated documentation). The LICENSE details I think are redundant as we already have the license specified in root of repository (and we've currently got MIT license here). Similar `authors` is likely to get out of date quickly - better to record this in metadata in `pyproject.toml`.

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+
```
matt-graham:
```suggestion
from pathlib import Path

```

For robust path handling.

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+    fab.update_environment(args)
+    fab.with_config(config)
+    fab.execute(fab.put_configs, config)
+
+    fab.env.neso_solver = solver
+
+    # This we presumably change somehow so that it gets changed throughout
+    # the SWEEP dir?
+    fab.env.neso_conditions_file = (
+        fab.find_config_file_path(config) + "/" + conditions_file_name
```
matt-graham:
```suggestion
        Path(fab.find_config_file_path(config)) / conditions_file_name
```
Ensures correct path separator used

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    # All of these should be in a config file somewhere
+    fab.env.neso_mesh_file = fab.find_config_file_path(config) + "/" + mesh_file_name
```
matt-graham:
```suggestion
    fab.env.neso_mesh_file = Path(fab.find_config_file_path(config)) / mesh_file_name
```

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+@fab.task
+def neso_ensemble(
+    config,
+    solver="views/gcc-hipsycl/bin/Electrostatic2D3V",
+    conditions_file_name="two_stream_conditions.xml",
+    mesh_file_name="two_stream_mesh.xml",
+    **args
+):
+    path_to_config = fab.find_config_file_path(config)
+    sweep_dir = path_to_config + "/SWEEP"
```
matt-graham:
```suggestion
    sweep_dir = Path(path_to_config) / "SWEEP"
```

templates/neso
```diff
@@ -0,0 +1,10 @@
+
+cd $job_results
+$run_prefix
+
+
+echo "Running NESO Electrostatic2D3V"
+
+echo ${neso_base_dir}$neso_solver $neso_conditions_file $neso_mesh_file
+
+${neso_base_dir}$neso_solver $neso_conditions_file $neso_mesh_file
```
matt-graham: Where is `neso_base_dir` specified?

templates/neso
```diff
@@ -0,0 +1,10 @@
+
+cd $job_results
+$run_prefix
+
+
+echo "Running NESO Electrostatic2D3V"
```
matt-graham:
```suggestion
echo "Running NESO solver $neso_solver"
```

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+    import base.fab as fab
+
+
+fab.add_local_paths("FabNeso")
+
+
+@fab.task
+def neso(
+    config,
+    solver="views/gcc-hipsycl/bin/Electrostatic2D3V",
```
matt-graham: If going the route of not specifying the full absolute path to the solver I would say it would be better to just have solver be the actual solver executable name, for example `Electrostatic2D3V`, with we just then needing for the user to point use to the relevant `bin` directory in which the solvers binaries have been built to as a config option or adding this directory to their path. This avoids baking in to the default an assumption that the build tree follow the Spack layout (which might not be the case if building directly with CMake) and will give a more user friendly command.

FabNeso.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # This we presumably change somehow so that it gets changed throughout
+    # the SWEEP dir?
+    fab.env.neso_conditions_file = (
+        fab.find_config_file_path(config) + "/" + conditions_file_name
+    )
+    # All of these should be in a config file somewhere
+    fab.env.neso_mesh_file = fab.find_config_file_path(config) + "/" + mesh_file_name
+
+    fab.job(dict(script="neso", wall_time="0:15:0", memory="2G"), args)
```
matt-graham: Do we need to specify `wall_time` and `memory` here? Both look a bit on the low side and we would probably want to make user overridable

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+):
+    """Use a dictionary with each parameter's high and low to create
+    a multi-dimensional sweep directory"""
+
+    print(
+        "{0} parameters found, requiring {1} divisions per parameter. "
+        "Creating {2} configurations...".format(
+            len(parameter_dict.keys()), n_dirs, n_dirs ** len(parameter_dict.keys())
+        )
+    )
```
matt-graham:
```suggestion
    print(
        f"{len(parameters_dict)} parameters found, "
        f"requiring {n_dirs} divisions per parameter. "
        f"Creating {n_dirs ** len(parameter_dict)} configurations..."
    )
```
[Formatted string literals (f-strings)](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals) are generally recommended over the `str.format` method in recent Python versions, as they're usually more readable and less likely to get bugs from putting arguments in wrong order.

The length of a `dict` is equivalent to the length of it's keys (or values) so I would go with the slightly shorter form.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # Calculate all of the parameter values and put them in a dict
+    scan_points = {}
+    list_all = []
+    for parameter in parameter_dict.keys():
+        param_value = parameter_dict[parameter][0]
+        scan_points[parameter] = []
+        for i in range(n_dirs):
+            scan_points[parameter].append(param_value)
+            list_all.append("{0}_{1}".format(parameter, i))
```
matt-graham:
```suggestion
            list_all.append(f"{parameter}_{i}")
```

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+    if destructive and os.path.isdir(sweep_path):
+        shutil.rmtree(sweep_path)
+    # Check whether the copy directory exists
+    copy_files = os.path.isdir(copy_dir)
+    # Now loop over each combination, create a directory, copy the
+    # configs and conditions, and edit the configuation to contain
+    # the correct parameters.
+    for comb in all_configs:
+        dirName = "-".join(comb)
+        dir_path = "{0}/SWEEP/{1}".format(sweep_path, dirName)
```
matt-graham:
```suggestion
        dir_name = "-".join(comb)
        dir_path = Path(sweep_path) / "SWEEP" / dir_name
```
Will also need to add an `from pathlib import Path`

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+    newData = []
+    for line in data:
+        if param in line:
+            line = line.split("=")[0] + "= {0} </P>\n".format(val)
+        newData.append(line)
+
+    with open(in_file_name, "w") as outFile:
+        for line in newData:
+            outFile.writelines(line)
```
matt-graham:
```suggestion
    new_data = []
    for line in data:
        if param in line:
            line = line.split("=")[0] + f"= {val} </P>\n"
        new_data.append(line)

    with open(in_file_name, "w") as out_file:
        for line in new_data:
            out_file.writelines(line)
```

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def edit_parameter(in_file_name, param, val):
+    """Edit a single parameter in the configuration
+    file to the desired value"""
+
+    print("Edit {1} : {0}".format(val, param))
+    data = []
+    with open(in_file_name, "r") as inFile:
+        data = inFile.readlines()
```
matt-graham:
```suggestion
    print(f"Edit {val} : {param}")
    data = []
    with open(in_file_name, "r") as in_file:
        data = in_file.readlines()
```

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
+""" Run ther create directory methods of the enseble_tools module
+"""
```
matt-graham: Some typos here. Also ideally the docstring should describe what the script does in a human readable form - for example similar to your `description` argument to `argparse.ArgumentParser`

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
...
+        help="An input dict of the parameters to be scanned.",
+        default="",
+    )
+    parser.add_argument("--dir_prefix", help="Prefix to call output dir", default="d")
+    args = parser.parse_args()
+
+    if args.parameter_dict:
+        # If this has been specified, we'll automatically create
+        # the directory tree as a multidimensional scan of these points
+        parameter_dict = eval(args.parameter_dict)
```
matt-graham:
```suggestion
        parameter_dict = eval(args.parameter_dict)
```
[Avoid `eval` wherever possible](https://stackoverflow.com/a/1832957). For the specific case here [`ast.literal_eval` is probably a safer bet](https://docs.python.org/3/library/ast.html#ast.literal_eval) or possibly [`json.loads`](https://docs.python.org/3/library/json.html#json.loads). I would also probably check that `parameter_dict` does end up being a `dict` instance and throw a `ValueError` or similar if not.

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
...
+        create_dict_sweep(
+            args.sweep_path,
+            args.n_divs,
+            args.destructive,
+            args.copy_dir,
+            args.edit_file,
+            parameter_dict,
+        )
+
+        sys.exit(0)
```
matt-graham:
```suggestion
```

Generally no need to call `sys.exit` as scripts will exit by default unless run in interactive mode, and we typically want to allow the user to run in interactive mode without forcing an exit.

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
...
+from ensemble_tools import create_dir_tree, create_dict_sweep
+
+import argparse
+import sys
+
+
+def main():
+    # Make the argument parser
+    parser = argparse.ArgumentParser(
+        prog="makeSweepDir", description="Makes a sweep directory for FabNeso"
```
matt-graham:
```suggestion
        description="Makes a sweep directory for FabNeso"
```

Better to leave `prog` as the default (which will correspond to name of script being called) to avoid this getting out of sync with filename as this has.

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
...
+        default="plugins/FabNeso/config_files/toCopy/",
+    )
+    parser.add_argument(
+        "--edit_file",
+        help="Template a parameter in this file",
+        default="two_stream_conditions.xml",
+    )
+    parser.add_argument(
+        "--para_to_template",
+        help="The parameter in the config " "file to template for the scan",
```
matt-graham:
```suggestion
        help="The parameter in the config file to template for the scan",
```

utils/make_sweep_dir.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument(
+        "--destructive",
+        "-d",
+        help="Deletes the previous tree if it already exists",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--copy_dir",
+        help="Copy contents of this dir to the sweep dirs",
+        default="plugins/FabNeso/config_files/toCopy/",
```
matt-graham: It looks this make an assumption about the working directory from which the script is called from for this to be a valid relative path which is a bit brittle. Also again typically should avoid hardcoding path separators to keep things agnostic to the operating system by using `Path` from `pathlib`. You can get the path of the current Python script / module using the special `__file__` attribute and could do something like `Path(__file__).parent / "plugins" / "FabNeso" / "config_files" / "toCopy"`

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+    for param_name in param_dict.keys():
+        edit_parameter(in_file_name, param_name, param_dict[param_name])
+
+
+def edit_parameter(in_file_name, param, val):
+    """Edit a single parameter in the configuration
+    file to the desired value"""
+
+    print("Edit {1} : {0}".format(val, param))
```
matt-graham:
```suggestion
    print(f"Edit {val} : {param}")
```

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def make_directory(directory_name, destructive):
+    """Make the directory tree at directory_name."""
+
+    if os.path.isdir(directory_name):
+        if destructive:
+            shutil.rmtree(directory_name)
+        else:
+            print("Path already exists and " "not in destructive mode! Returning")
```
matt-graham:
```suggestion
            print("Path already exists and not in destructive mode! Returning")
```

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+        if edit_file:
+            encode_conditions_file(dir_path + "/" + edit_file, parameters)
+    return 0
+
+
+def copy_dir_contents(dir_path, copy_dir):
+    """Copy the contents of copy_dir -> dir_path"""
+
+    for f in os.listdir(copy_dir):
+        shutil.copy(copy_dir + f, dir_path)
```
matt-graham: I think [`shutil.copytree`](https://docs.python.org/3/library/shutil.html#shutil.copytree) already does what you want here.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+        # Create parameter map for conditions encoding
+        parameters = {}
+        for parameter in parameter_dict.keys():
+            param_index = [item for item in comb if item.startswith(parameter)][
+                0
+            ].split("_")[-1]
+            parameters[parameter] = scan_points[parameter][int(param_index)]
+        if edit_file:
+            encode_conditions_file(dir_path + "/" + edit_file, parameters)
+    return 0
```
matt-graham:
```suggestion
```

Assuming this is being returned as a status code, this is generally unnecessary and non-idiomatic in Python. A function which doesn't have an explicit return value in Python will just return `None` by default and functions are generally assumed to have run without issues unless an exception is raised. If you were to return some sort of status code you would probably also want to use `bool` literals rather than 0 or 1.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+
+def make_directory(directory_name, destructive):
+    """Make the directory tree at directory_name."""
+
+    if os.path.isdir(directory_name):
+        if destructive:
+            shutil.rmtree(directory_name)
+        else:
+            print("Path already exists and " "not in destructive mode! Returning")
+            return 0
```
matt-graham: It would be more idiomatic to raise an exception here for example
```Python
msg = "Path already exists and not in destructive mode."
raise FileExistsError(msg)
```
and then when calling the function just assume it runs correctly unless an exception is raised which will cause script to exit unless explicitly handled with a `try... except`.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,158 @@
...
+    os.makedirs(directory_name)
+    return 1
```
matt-graham:
```suggestion
```
Similar to above generally returning integer status codes is not recommended in Python as it's unclear what they represent.

utils/ensemble_tools.py
matt-graham: Also `writelines` accepts a list of lines so don't think you need the inner loop here.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,159 @@
...
+def encode_conditions_file(in_file_name, param_dict):
+    """Encode a configuration file with a dict of input values"""
+
+    for param_name in param_dict.keys():
+        edit_parameter(in_file_name, param_name, param_dict[param_name])
+
+
+def edit_parameter(in_file_name, param, val):
+    """Edit a single parameter in the configuration
+    file to the desired value"""
```
matt-graham: The approach here for changing parameter values is very brittle. For example I think if the `param` string appears anywhere else in the file other than the parameter definition we'll still end up substituting this with something, which will probably break things. I would say either better to use a combination of parsing XML and querying for relevant node, and then using a regular expression or similar to parse the parameter definition as in https://github.com/UCL/neso-calibration/blob/52e63174f6347d501f7f62771107949be4e299d4/src/nesopy/model.py#L121 or using a templated conditions file would be better from a robustness and flexibility perspective.

matt-graham (changes requested):

templates/neso
dleggat: neso_base_dir is defined in fabsim/deploy/machines_user.yml

This approach was recommended in the FabSIM documentation

FabNeso.py
dleggat: The header comment was written automatically when making the plugin and is consistent with the other, existing plugins. I'm happy to change it in the way you describe, just noting that this is another one of those 'default FabSIM' things!

FabNeso.py
dleggat: I could see this being another part of the machines_user.yml configuration, like the neso_base_dir, so that it can be set per machine.

dleggat: And in fact it only gets used in association with neso_base_dir, so if that gets changed to, say, neso_bin as the environment variable, it has the same effect while making things more user friendly

dleggat: And now implementing this I've just remembered the reservation here, that by pre-defining the bin directory you're already choosing the compilation type - i.e. the oneapi and gcc compilations end up in separate directories, so we're pre-supposing which one is to be used here. Maybe that's fine, though?

FabNeso.py
dleggat: Moved to be optional inputs to the task

FabNeso.py
matt-graham: Yeah sorry realised this is from FabSIM side rather than something you've added. It looks like in general their code is a bit out of date with current Python standards. While there is some value in keeping with the existing conventions in a project, in this case I think there are enough downsides (using an inline comment will mean no `__doc__` attribute is set, having an author list and licence details in comment makes it difficult to ensure these are kept up to date) that I would err with breaking their convention.

templates/neso
matt-graham: Shouldn't this be defined in a file local to this repository though as `fabsim/deploy/machines_user.yml` is outside of this repository? From the docs at https://fabsim3.readthedocs.io/en/latest/create_new_plugin/ it looks like we probably want a `machines_fabneso_user.yml` file in the root of the repository where any local environment variables are defined.

FabNeso.py
matt-graham: I think we want the user to be able to specify the relevant `bin` directory to this doesn't bake in an assumption of a particular compiler toolchain like you say (or using Spack at all) but I don't think this needs to necessarily be configurable as a command line argument, but instead would be better as an environment variable. So `solver` just corresponds to the name of the solver executable and elsewhere we have a configurable variable defining location of where executables are.

FabNeso.py
dleggat: Weirdly, the sweep_dir is required to be a string - fabSIM crashes if it's left as a Path. So I've cast it back to str() in this line after the concatenation has happened. This hasn't been a problem with any of the other variables (yet...).

templates/neso
dleggat: I've updated the plugin to read from a local machines_FabNeso_user.yml file if such a file exists, and moved the base directory environment variable to there.

In the repository, I've included a machines_FabNeso_user_example.yml rather than the direct file, as this appears to be the FabSIM standard in the other plugins. FabNeso doesn't actually read this file, though, so the user has to change it locally. Happy to change this if you think the actual file would be better.

FabNeso.py
dleggat: I ended up updating neso_base_dir to neso_bin_dir in the newly created local environment variable file, which contains this information instead. The base_dir is not used elsewhere now that we're specifying the mesh and conditions files, so this change made sense.

FabNeso.py
dleggat: This is another candidate for inclusion in the local environment variable definition - then we can specify these separately depending on the machines we're running on

FabNeso.py
dleggat: I've gone ahead and updated these

utils/ensemble_tools.py
dleggat: I was a little worried about this method too - I realised that, for example, the hw.xml has an 's' parameter that would certainly lead to undesirable consequences if we were to try and template it.

I've implemented a version for pattern matching the XML parsed parameters using a slightly adjusted version of the method above. I'm using lxml because it preserves comments in the XML file by default.

FabNeso.py
```diff
@@ -0,0 +1,82 @@
...
+
+    fab.env.neso_solver = solver
+
+    # This we presumably change somehow so that it gets changed throughout
+    # the SWEEP dir?
+    fab.env.neso_conditions_file = (
+        Path(fab.find_config_file_path(config)) / conditions_file_name
+    )
+    # All of these should be in a config file somewhere
+    fab.env.neso_mesh_file = Path(fab.find_config_file_path(config)) / mesh_file_name
```
matt-graham: After installing and trying this out locally I'm not sure these paths are set up as we'd need to. As far as I can tell the previous `fab.execute(fab.put_configs, config)` call copies the subdirectory of `config_files` named `config` to a directory on the machine being targetted (for now `localhost`). As this may be a remote machine I think the intention is that any paths we use in the templated script we run are from the locally copied `config` directory. This  sets `neso_mesh_file` and `neso_conditions_file` to point to the source configuration files rather than the local copy (and on a remote machine these paths won't exist).

utils/ensemble_tools.py
matt-graham: Using `lxml` is fine though [it looks it is may be possible to use `xml` module in standard library and preserve comments](https://stackoverflow.com/a/66668127). We will need to add a `requirements.txt` to the root of the repository with `lxml` listed if using to ensure it is installed when the plug-in is.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+    try:
+        make_directory(sweep_path, destructive)
+    except FileExistsError:
+        return
+
+    # Set the initial value of the scanned parameter to the lower limit
+    para_val = scan_range[0]
+
+    for i in range(n_dirs):
+        new_dir = Path(sweep_path) / "SWEEP" / outdir_prefix + i
```
matt-graham:
```suggestion
        new_dir = Path(sweep_path) / "SWEEP" / f"{outdir_prefix}{i}"
```
Currently raises `TypeError: unsupported operand type(s) for +: 'PosixPath' and 'int'`

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+
+    copy_files = os.path.isdir(copy_dir)
+    if not copy_files:
+        print("No correct copy dir found, just creating the tree for now!")
+
+    # Make the base directory
+    try:
+        make_directory(sweep_path, destructive)
+    except FileExistsError:
+        return
```
matt-graham: I would leave the exception unhandled rather than returning without doing anything as its not clear to user that something went wrong otherwise and the error message should be informative

```suggestion
    make_directory(sweep_path, destructive)
```

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+        new_dir = Path(sweep_path) / "SWEEP" / outdir_prefix + i
+        # Make the directory
+        os.makedirs(new_dir)
+        # If we're copying files, do so
+        if copy_files:
+            copy_dir_contents(new_dir, copy_dir)
+        # Now we edit the parameter file for our
+        # template scan if we're doing that
+        if os.path.isfile(new_dir + edit_file) and parameter_to_scan:
+            edit_parameter(new_dir + edit_file, parameter_to_scan, para_val)
```
matt-graham:
```suggestion
        if (new_dir / edit_file).is_file() and parameter_to_scan:
            edit_parameter(new_dir / edit_file, parameter_to_scan, para_val)
```
`pathlib.Path` instances should be joined with `/` (they don't support string concatenation with `+`). Also more idiomatic to use `Path` methods rather than `os.path` functions

matt-graham: Also I would say we might want to raise an exception if `(new_dir / edit_file).is_file()` is false or `parameter_to_scan` is `None` rather than only call `edit_parameter` if the inverse is true as otherwise we are just creating a sweep directory with verbatim copies of the contents of `copy_dir` which seems to be unlikely what the user would want. Also generally better from a readability and robustness perspective to be explicit when testing if value is empty or `None` rather than relying on empty objects being considered false - so for example `parameter_to_scan != ""` or `parameter_to_scan is not None` rather than just `parameter_to_scan`.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+        if match is None:
+            # This happens for the constructed elements, so skip them
+            continue
+        key = match.group("key")
+        if key == param:
+            new_element = ET.Element("P")
+            new_element.text = f" {param} = {val} "
+            parameters.insert(parameters.index(element), new_element)
+            parameters.remove(element)
```
matt-graham:
```suggestion
            element.text = f" {param} = {val} "
```
Looks like editing `element.text` in place works from a quick test - not sure why I didn't follow this pattern in `nesopy`!

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+    root = data.getroot()
+    parameters = root.find("CONDITIONS").find("PARAMETERS")
+    for element in parameters.iter("P"):
+        match = re.match(
+            r"\s*(?P<key>\w+)\s*=\s*(?P<value>[0-9.-]+)\s*",
+            element.text,
+        )
+        if match is None:
+            # This happens for the constructed elements, so skip them
+            continue
```
matt-graham: Adding a `breakpoint()` here and running `make_sweep_dir.py` script locally suggests this if-block isn't entered in normal operation. What does 'constructed elements' in the comment here refer to? If we don't get a match I would be inclined to raise an error as that suggests we have a parameter defnition in a format we don't know how to handle.

utils/ensemble_tools.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def encode_conditions_file(in_file_name, param_dict):
+    """Encode a configuration file with a dict of input values"""
+
+    for param_name in param_dict.keys():
+        edit_parameter(in_file_name, param_name, param_dict[param_name])
+
+
+def edit_parameter(in_file_name, param, val):
```
matt-graham: If rather than passing a singe `param`, `val` pair to `edit_parameter` you instead passed a parameter to value dict then you could parse the XML file only once - something like

```Python
def edit_parameters(in_file_name, parameter_overrides):
    """Edit one or more parameters in conditions file."""
    data = ET.parse(in_file_name)
    root = data.getroot()
    parameters = root.find("CONDITIONS").find("PARAMETERS")
    for element in parameters.iter("P"):
        match = re.match(
            r"\s*(?P<key>\w+)\s*=\s*(?P<value>[0-9.-]+)\s*",
            element.text,
        )
        if match is None:
            raise ValueError(f"Invalid parameter specification in base file: {element.text}")
        key = match.group("key")
        if key in parameter_overrides:
            element.text = f" {key} = {parameter_overrides[key]} "
    ET.indent(root, space="    ")
    data.write(in_file_name)
```
which would then mean we could get rid of `encode_conditions_file` and use `edit_parameters` in place of both the current `encode_condition_file` and `edit_parameter` calls.

utils/ensemble_tools.py
dleggat: I've used the method linked, so have gone back to the regular ```xml ``` package.

utils/ensemble_tools.py
matt-graham: Ah I think I understand now - by 'constructed elements' you mean parameter definitions which involve other parameter values rather such as in the `hw.xml` file? Here we can probably just use a looser regex that will match these cases as well, particularly as we don't actually need to capture the actual original values here in a group (I did this in `nesopy` so we could record a dictionary of all parameter values including those not overriden). We only need to capture the key in a group so something like `r"\s*(?P<key>\w+)\s*="`  should work I think.

FabNeso.py
dleggat: I've just taken out the ```fab.find_config_file_path()``` methods, and now I believe it has the desired behaviour. I'd misunderstood this method as finding the job directory after the configs had been transferred, rather than the original location of the configs.

matt-graham: Yeah I had also initially assumed that was what `fab.find_config_file_path()` did.

utils/ensemble_tools.py
dleggat: Changed these to raise exceptions as suggested

utils/ensemble_tools.py
dleggat: It was indeed from those parameters in the ```hw.xml```. Changing the regex does indeed fix the problem, so I've changed that now.

utils/ensemble_tools.py
dleggat: I noticed this too when I was looking through the XML with comments example you sent. so had already updated it this way!

matt-graham (approved):

PR #15, opened 2023-11-24
Allow command line arguments to override configuration input parameters for a single instance of running NESO

### LINKED ISSUE #11: Allow overriding parameter values in `neso` task
Currently the `neso` task only allows running solvers with pre-defined conditions file in the `config_files` directory. It would be useful to add an option to override parameters at the command-line. This should be possible to do by creating a (temporary) copy of the config directory before these lines
https://github.com/UCL/fabneso/blob/89bba5ca73c333cf16e8b8dddee164a67b1eb867/FabNESO.py#L35-L36
and using the `edit_parameters` function in `utils/ensemble_tools.py`. I think we can capture arbitrary keyword arguments to the task by adding `**kwargs` to the argument list, and then just pass these directly as the parameter dictionary. This would preclude parameter names that clash with any of the other arguments to the task, but it doesn't seem likely this will be much of an issue.

dleggat:
Addressing https://github.com/UCL/FabNESO/issues/11

### DIFF
--- a/FabNESO.py
+++ b/FabNESO.py
@@ -40,7 +40,7 @@ def neso(
         **kwargs: Additional kwargs will be passed to ensemble_tools.edit_parameters to
                   create a temporary conditions file with these overriden parameters
     """
-    # Use a temporary directory context so that we can handle parsameter inputs
+    # Use a temporary directory context so that we can handle parameter inputs
     # from the command line
     original_config_path = Path(fab.find_config_file_path(config))
     temporary_context: Union[TemporaryDirectory, nullcontext] = (
--- a/FabNESO.py
+++ b/FabNESO.py
@@ -5,6 +5,8 @@ Defines tasks for running simulations using Neptune Exploratory Software (NESO).
 """
 
 from pathlib import Path
+from plugins.FabNESO.utils.ensemble_tools import edit_parameters
+import shutil
 
 try:
     import fabsim.base.fab as fab
@@ -22,6 +24,7 @@ def neso(
     solver: str = "Electrostatic2D3V",
     conditions_file_name: str = "conditions.xml",
     mesh_file_name: str = "mesh.xml",
+    **kwargs
 ):
     """
     Run a single NESO solver instance.
@@ -31,7 +34,14 @@ def neso(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
+        **kwargs: Additional kwargs will be passed to ensemble_tools.edit_parameters to create a temporary conditions file with these overriden parameters
     """
+    #If there have been additional parameters provided, create a copy of the conditions file and edit the input parameters
+    if not kwargs == {}:
+        shutil.copytree(fab.find_config_file_path(config),Path(fab.find_config_file_path(config)).parent / "tmp_configs")
+        config = "tmp_configs" #switch our config to the new tmp ones
+        edit_parameters(Path(fab.find_config_file_path(config)) / conditions_file_name, kwargs)
+            
     fab.with_config(config)
     fab.execute(fab.put_configs, config)
     fab.job(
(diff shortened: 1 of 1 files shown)


FabNESO.py
```diff
@@ -31,7 +34,21 @@ def neso(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
+        **kwargs: Additional kwargs will be passed to ensemble_tools.edit_parameters to
+                  create a temporary conditions file with these overriden parameters
     """
+    # If there have been additional parameters provided, create a copy of the
```
matt-graham: At the moment if there is an error raised after the temporary conditions directory is created but before the final `shutil.rmtree` call then we will end up with a `tmp_configs` directory remaining on disk which will cause an error next time we try to run with a parameter override unless we manually clean up. 

We can use [the ability of the `tempfiles.TemporaryDirectory` class to be used as a context manager](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) here to make sure that the temporary directory will always be cleaned up even when an exception is raised, and also ensure it always gets a unique name, for example something like

```Python
    config_path = Path(fab.find_config_file_path(config))
    with (
        TemporaryDirectory(prefix=f"{config}_", dir=config_path.parent)
        if kwargs != {} else nullcontext()
    ) as temporary_config_directory:
        if kwargs != {}: 
            # If there have been additional parameters provided, create a copy of the
            # conditions file and edit the input parameters
            temporary_config_path = Path(temporary_config_directory)
            shutil.copytree(config_path, temporary_config_path, dirs_exist_ok=True)
            edit_parameters(temporary_config_path / conditions_file_name, kwargs)
            config = temporary_config_path.name  
        fab.with_config(config)
        fab.execute(fab.put_configs, config)
        fab.job(
            {
                "script": "neso",
                "neso_solver": solver,
                "neso_conditions_file": conditions_file_name,
                "neso_mesh_file": mesh_file_name,
            }
        )
```

with 

```Python
from contextlib import nullcontext
from tempfile import TemporaryDirectory
```

added to the imports.

matt-graham:
> Addressing #11

Also just FYI, to link an issue to a pull request automatically you can use [a few different keywords (`closes`, `fixes`, `resolves` plus conjugation variations)](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

FabNESO.py
```diff
@@ -31,17 +37,38 @@ def neso(
...
-            "neso_mesh_file": mesh_file_name,
-        }
+    # Use a temporary directory context so that we can handle parsameter inputs
+    # from the command line
+    original_config_path = Path(fab.find_config_file_path(config))
+    temporary_context: Union[TemporaryDirectory, nullcontext] = (
+        TemporaryDirectory(prefix=f"{config}_", dir=original_config_path.parent)
+        if not kwargs == {}
+        else nullcontext()
     )
```
dleggat, line 50: I had to add this little explicit pre-definition of the context to satisfy mypy. It was complaining during the with statement that it couldn't find the __enter__ and __exit__ consistently when there were differing types during the definition.

FabNESO.py
```diff
@@ -31,17 +37,38 @@ def neso(
...
-    fab.with_config(config)
-    fab.execute(fab.put_configs, config)
-    fab.job(
-        {
-            "script": "neso",
-            "neso_solver": solver,
-            "neso_conditions_file": conditions_file_name,
-            "neso_mesh_file": mesh_file_name,
-        }
+    # Use a temporary directory context so that we can handle parsameter inputs
```
matt-graham:
```suggestion
    # Use a temporary directory context so that we can handle parameter inputs
```

matt-graham (approved):
Looks good to merge to me other than one small typo in comment

PR #19, opened 2023-11-29
Restructure plugin as standalone package

### LINKED ISSUE #9: Figure out how to structure project and specify plug-in dependencies
## How FabSIM installs and loads plug-ins

When [installing a plug-in FabSIM clones the whole plug-in repository](https://github.com/djgroen/FabSim3/blob/d94fb1da49e732719cbf7b22bb22e970367fca2e/fabsim/base/setup_fabsim.py#L64-L75) to a subdirectory `plugins/{plugin_name}` under the top-level `FabSim3` repository directory, where `{plugin_name}`  is [the key for the plug-in in the `fabsim/deploy/plugins.yml` file](https://github.com/djgroen/FabSim3/blob/master/fabsim/deploy/plugins.yml). So for example if we assume the `plugin_name = "fabneso"` here then the contents of this repository will be cloned to `plugins/fabneso`.

The `plugins` directory [contains an (empty) `__init__.py`](https://github.com/djgroen/FabSim3/blob/master/plugins/__init__.py) so will be recognised by Python as a package. During setup FabSIM3 also [asks the user to add the root FabSIM3 repository directory to both `PYTHONPATH` and `PATH` environment variables](https://github.com/djgroen/FabSim3/blob/d94fb1da49e732719cbf7b22bb22e970367fca2e/configure_fabsim.py#L231-L239) which means that the `plugins` package will be importable as will any sub-packages / modules within it. 

When loading a plug-in FabSIM3 [uses `importlib` to import the module `plugins.{plugin_name}.{plugin_name}`](https://github.com/djgroen/FabSim3/blob/d94fb1da49e732719cbf7b22bb22e970367fca2e/fabsim/deploy/machines.py#L240). It then does some slightly dubious looking manipulation of the `globals` dict of the calling function to add all top-level names defined in `plugins.{plugin_name}.{plugin_name}` (either as explicitly specified by `__all__` attribute or any names defined in modules `__dict__` attributed which don't begin with `_`) to the calling functions `globals` namespace.

## What does this all mean for us?

I *think* based on the above we could have a local `fabneso` directory which we make a standard Python package by including an `__init__.py` file. We could then have a series of modules in this package, for example `utils.py`, `tasks.py` defining the plug-in functionality, with the `__init__.py` then locally importing and exposing any names from these other modules which we wish to expose as the plugin's public interface. This is slightly odd in that the `plugins/fabneso` directory which we clone in to won't have an `__init__.py` and be a standard Python package, however it should still be importable as [a namespace (sub)package](https://docs.python.org/3/reference/import.html#namespace-packages) (at least I think this is how it would work) and follows the pattern used by other FabSIM3 plug-ins which don't seem to include a top-level `__init__.py` file.

Structuring like this would then have the advantage that we can also install the plug-in as a standard Python package by including a `pyproject.toml` metadata file (or `setup.py` script) and not need to do too much hacking to be able to use `tox` and `pytest` to run unit tests in a `tests` directory.  One slightly complicating issues is whether to consider FabSIM3 itself a dependency and how to specify this as it cannot be installed using `pip` as far as I'm aware; ideally we would so that `ruff` and `mypy` can do checks on imports from `fabsim` namespace.

FabSIM3 also allows specifying any plug-in specific Python dependencies in a `requirements.txt` file in the root of the repository, which [is passed to a `pip install -r` command when installing the plug-in](https://github.com/djgroen/FabSim3/blob/d94fb1da49e732719cbf7b22bb22e970367fca2e/fabsim/base/setup_fabsim.py#L80-L86). We can use the ability to [reference an external `requirements.txt` as a dynamic `project.dependencies` property in the `pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) to have this also automatically picked up when installing the package locally, to ensure we only need to specify dependencies in one place.

matt-graham:
Resolves #9 

Restructures current `FabNESO.py` module in to an equivalently named package `FabNESO` with `__init__.py` set to import the tasks to be exposed from a new module `FabNESO/tasks.py` with the previous content from `FabNESO.py`. This allows using standard relative imports to import names from other modules such as `ensemble_tools` rather than relying on FabSim's more brittle approach of directly manipulating `PYTHONPATH`.

A `pyproject.toml` metadata file describing package has been added which also allows the package to be directly installed, for example by running `pip install -e .` (to install in editable mode) from root of repository. This also will also allow using `tox` to run tests in isolated environments.

The `pyproject.toml` is based on the cookie cutter template at https://github.com/UCL-ARC/python-tooling/ which uses a stricter Ruff configuration than the default, with all rules included and then a few specific rules excluded. To get the `pre-commit` checks to pass some fixes were needed (some automatically done by Ruff, others requiring some manual editing). I have a added a specific global ignore for [Ruff rule N999 (`invalid-module-name`)](https://docs.astral.sh/ruff/rules/invalid-module-name/) (which doesn't like the capitalization in `FabNESO` package name) as we are intentionally following FabSim3's existing convention here of plugin names being in `CamelCase`. I also increased the maximum allowed arguments to a function from the default of 5 to 10.

I also added a couple of local ignores for [Ruff rule S314 (`suspicious-xml-element-tree-usage`)](https://docs.astral.sh/ruff/rules/suspicious-xml-element-tree-usage/) - we could consider using the safer `defusedxml` here but as we can probably assume the XML files being parsed here are to be trusted, and this would require have non standard library dependency, for now I've kept with the current approach.


dleggat (approved):
Everything seems to work fine. Had an issue installing when I first tried, but it turned out I needed to update pip, and once that was done it has worked fine.

PR #23, opened 2023-11-30
First draft of tests on the ensemble_tools

### LINKED ISSUE #12: Adding tests for utility functions
We should add a Pytest test suite for at a  minimum the utility functions and add a GitHub Actions workflow to automate running these on pull-requests and pushes to `main`.

We ideally would also test the plug-in tasks as well, but this might be a bit more challenging, both because of the difficulty in installing FabSIM in an automated way and because of the need to have some form of proxy for the NESO solver executables (though we could just use a script which e.g. touches some output files without actually running anything).

dleggat:
A first draft of the tests for ensemble_tools.

The tests use /tmp/ to create dummy sweep directories with a range of templated parameters, and check the directory structure and xml output for validity.

The tests work with `pytest`, but do not currently run in `tox`. I get the following error:

```
=========================================================================================================================================================== ERRORS ===========================================================================================================================================================
_______________________________________________________________________________________________________________________________________ ERROR collecting tests/test_ensemble_tools.py ________________________________________________________________________________________________________________________________________
ImportError while importing test module '/home/duncan/Work/aquifer/fabnesoPlugin/paraOverride/tests/test_ensemble_tools.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
FabNESO/tasks.py:13: in <module>
    from fabsim.base import fab
E   ModuleNotFoundError: No module named 'fabsim'

During handling of the above exception, another exception occurred:
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_ensemble_tools.py:13: in <module>
    from FabNESO.ensemble_tools import create_dict_sweep, create_dir_tree, edit_parameters
FabNESO/__init__.py:3: in <module>
    from .tasks import neso, neso_ensemble
FabNESO/tasks.py:15: in <module>
    from base import fab
E   ModuleNotFoundError: No module named 'base'
```

Which I'm assuming means something in the .toml needs to be changed to allow pytest access to the base fabsim classes, but adding in the fabsim pythonpath argument to the pytest init hook hasn't solved the problem as I hoped it would.

Resolves #12

### DIFF
--- a/FabNESO/__init__.py
+++ b/FabNESO/__init__.py
@@ -1,8 +1,14 @@
 """Neptune Exploratory SOftware (NESO) plugin for FabSim3."""
-import contextlib
 
-with contextlib.suppress(ImportError):
+try:
     from .tasks import neso, neso_ensemble
 
+    __all__ = ["neso", "neso_ensemble"]
 
-__all__ = ["neso", "neso_ensemble"]
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        "Cannot import FabNESO tasks - fabsim.base.fab may not be importable",
+        stacklevel=2,
+    )
--- a/FabNESO/__init__.py
+++ b/FabNESO/__init__.py
@@ -1,0 +1,5 @@
+"""Neptune Exploratory SOftware (NESO) plugin for FabSim3."""
+
+from .tasks import neso, neso_ensemble
+
+__all__ = ["neso", "neso_ensemble"]
--- a/FabNESO/__init__.py
+++ b/FabNESO/__init__.py
@@ -1,5 +1,8 @@
 """Neptune Exploratory SOftware (NESO) plugin for FabSim3."""
+import contextlib
 
-from .tasks import neso, neso_ensemble
+with contextlib.suppress(ImportError):
+    from .tasks import neso, neso_ensemble
+
 
 __all__ = ["neso", "neso_ensemble"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
+    "defusedxml",
 ]
 description = "Neptune Exploratory SOftware (NESO) plugin for FabSim3"
 dynamic = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,129 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+    "setuptools",
+    "setuptools-scm",
+]
+
+[project]
+authors = [
+    {email = "arc.collaborations@ucl.ac.uk", name = "UCL ARC Collaborations team"},
+]
+classifiers = [
+    "Operating System :: POSIX",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Typing :: Typed",
+]
+dependencies = [
+]
+description = "Neptune Exploratory SOftware (NESO) plugin for FabSim3"
+dynamic = [
+    "version",
+]
+keywords = [
+]
+name = "FabNESO"
+optional-dependencies = {dev = [
+    "build",
+    "mypy",
+    "pre-commit",
+    "pytest",
+    "ruff",
+    "tox>=4",
+    "twine",
+]}
+readme = "README.md"
+requires-python = ">=3.10"
+license.file = "LICENSE.md"
+urls.homepage = "https://github.com/UCL/FabNESO"
+
+[tool.coverage]
+report = {skip_covered = true, sort = "cover"}
+run = {branch = true, parallel = true, source = [
+    "FabNESO",
+]}
+paths.source = [
+    "FabNESO",
+    ".tox*/*/lib/python*/site-packages",
+]
+
+[tool.mypy]
+explicit_package_bases = true
+plugins = [
+    "numpy.typing.mypy_plugin",
+]
+
+[tool.pytest.ini_options]
+addopts = "--color=yes -v"
+testpaths = [
+    "tests",
+]
+
+[tool.ruff]
+fix = true
+force-exclude = true
+ignore = [
+    "COM812", # trailing commas (ruff-format recommended)
+    "D203", # no-blank-line-before-class
+    "D212", # multi-line-summary-first-line
+    "D407", # removed dashes lines under sections
+    "D417", # argument description in docstring (unreliable)
+    "ISC001", # simplify implicit str concatenation (ruff-format recommended)
+    "N999", # invalid-module-name - follow FabSim3 convention of CamelCase plugin names
+]
+per-file-ignores = {"tests*" = [
+    "INP001",
+    "S101",
+]}
+select = [
+    "ALL",
+]
+target-version = "py310"
+isort.known-first-party = [
+    "FabNESO",
+]
+mccabe.max-complexity = 18
+pep8-naming.classmethod-decorators = [
+    "classmethod",
+]
+pylint.max-args = 10
+
+[tool.setuptools]
+packages = ["FabNESO"]
+
+[tool.setuptools_scm]
+local_scheme = "no-local-version"
+write_to = "FabNESO/_version.py"
+
+[tool.tomlsort]
+all = true
+spaces_indent_inline_array = 4
+trailing_comma_inline_array = true
+overrides."project.classifiers".inline_arrays = false
+overrides."tool.coverage.paths.source".inline_arrays = false
+
+[tool.tox]
+legacy_tox_ini = """
+    [gh-actions]
+    python =
+        3.10: py310
+        3.11: py311
+        3.12: py312
+
+    [testenv]
+    commands =
+        pytest --cov --cov-report=xml
+    deps =
+        pytest
+        pytest-cov
+
+    [tox]
+    env_list =
+        py310
+        py311
+        py312
+"""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
-    "defusedxml",
 ]
 description = "Neptune Exploratory SOftware (NESO) plugin for FabSim3"
 dynamic = [
--- a/tests/test_ensemble_tools.py
+++ b/tests/test_ensemble_tools.py
@@ -24,12 +24,12 @@ def test_edit_parameters(tmpdir: local) -> None:
 
     edit_parameters(temp_conditions_path, {parameter_to_test: parameter_test_value})
 
-    assert check_parameter_in_conditions(
+    assert _check_parameter_in_conditions(
         temp_conditions_path, parameter_to_test, parameter_test_value
     )
 
 
-def check_parameter_in_conditions(
+def _check_parameter_in_conditions(
     conditions_file_name: Path, parameter_name: str, expected_value: float
 ) -> bool:
     """Return True if parameter_name has approx expected_value."""
@@ -93,7 +93,7 @@ def test_create_dir_tree(tmpdir: local) -> None:
         para_value = scan_range[0] + (i / (n_dirs - 1)) * (
             scan_range[1] - scan_range[0]
         )
-        assert check_parameter_in_conditions(cond_file, parameter_to_scan, para_value)
+        assert _check_parameter_in_conditions(cond_file, parameter_to_scan, para_value)
     # Should raise an exception when trying to edit the mesh file
     edit_file = "mesh.xml"
     with pytest.raises(ValueError, match=r".* a CONDITIONS node.*"):
@@ -153,11 +153,11 @@ def test_create_dict_sweep(tmpdir: local) -> None:
             para_value = scan_range[0] + (indices[i] / (n_divs - 1)) * (
                 scan_range[1] - scan_range[0]
             )
-            assert check_parameter_in_conditions(
+            assert _check_parameter_in_conditions(
                 this_dir / "conditions.xml", parameter, para_value
             )
             with pytest.raises(ValueError, match=r".*CONDITIONS.*"):
                 # The mesh file should be not edited
-                assert check_parameter_in_conditions(
+                assert _check_parameter_in_conditions(
                     this_dir / "mesh.xml", parameter, para_value
                 )
--- a/tests/test_ensemble_tools.py
+++ b/tests/test_ensemble_tools.py
@@ -69,6 +69,30 @@ def _check_parameter_in_conditions(
     return (n_equals, n_different)
 
 
+def test_check_parameter_in_conditions(tmp_path: Path) -> None:
+    """Test that the private xml parser method of this class."""
+    parameter_to_test = "particle_initial_velocity"
+    parameter_test_value = 10.0
+    test_cond_path = tmp_path / "conditions.xml"
+    shutil.copyfile(
+        Path(__file__).parents[1] / "config_files" / "two_stream" / "conditions.xml",
+        test_cond_path,
+    )
+    edit_parameters(test_cond_path, {parameter_to_test: parameter_test_value})
+    # Test that a different value of the test_value returns exactly 1 non-matched result
+    n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
+        test_cond_path, parameter_to_test, parameter_test_value + 10.0
+    )
+    assert n_equal_in_value == 0
+    assert n_different_in_value == 1
+    # Pass a fake parameter name. Should find 0 results
+    n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
+        test_cond_path, "fake_parameter", parameter_test_value
+    )
+    assert n_equal_in_value == 0
+    assert n_different_in_value == 0
+
+
 @pytest.mark.parametrize("n_dirs", [1, 3, 5, 10])
 @pytest.mark.parametrize("destructive", [True, False])
 @pytest.mark.parametrize(
--- a/tests/test_ensemble_tools.py
+++ b/tests/test_ensemble_tools.py
@@ -8,7 +8,6 @@ from xml.etree import ElementTree
 
 import pytest
 
-# from plugins.FabNESO.utils.make_sweep_dir
 from FabNESO.ensemble_tools import create_dict_sweep, create_dir_tree, edit_parameters
 
 
@@ -68,15 +67,29 @@ def _check_parameter_in_conditions(
     return (n_equals, n_different)
 
 
-def test_create_dir_tree(tmp_path: Path) -> None:
+@pytest.mark.parametrize("n_dirs", [1, 3, 5, 10])
+@pytest.mark.parametrize("destructive", [True, False])
+@pytest.mark.parametrize(
+    "parameter_and_range",
+    [
+        ["particle_initial_velocity", -4.0, 5.0],
+        ["particle_charge_density", 101.0, 108.0],
+        ["particle_number_density", 101.0, 108.0],
+    ],
+)
+def test_create_dir_tree(
+    tmp_path: Path,
+    *,
+    n_dirs: int,
+    destructive: bool,
+    parameter_and_range: list,
+) -> None:
     """Test the create_dir_tree method of ensemble_tools."""
     test_sweep_dir = tmp_path / "test_sweep"
-    n_dirs = 5
-    destructive = True
     copy_dir = Path("config_files") / "two_stream"
     edit_file = "conditions.xml"
-    parameter_to_scan = "particle_initial_velocity"
-    scan_range = (0.2, 5.0)
+    parameter_to_scan = parameter_and_range[0]
+    scan_range = (parameter_and_range[1], parameter_and_range[2])
     outdir_prefix = "asd"
 
     create_dir_tree(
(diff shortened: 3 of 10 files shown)


matt-graham:
Ah yes this is going to be a pain. I think the root of the issue that `tox` by design creates a new isolated environment and install the package and its dependencies in it before running tests (or other commands). So while when you run `pytest` directly from an environment in which FabSim3 has been installed it will be able to import successfully, but running `pytest` from within `tox`, the manual manipulations of `PYTHONPATH` than the FabSim3 install performs do not have any effect and the isolated environment won't be able to import from `fabsim`.

There are a few routes we could go
1. Make the import from `.tasks` in `FabNESO/__init__.py` optional by wrapping in some `try: ... except ImportError:` logic. This will then allow Pytest to import `FabNESO` even when `fabsim` cannot be imported, and if we are only testing the `ensemble_tools` module for now, should suffice.
2. See if we can manually have a FabSim install made available within the isolated `tox` environment by manipulating `PYTHONPATH` in the `tox` environment definition. This is likely to be brittle and will require having FabSim3 set up in advance before being able to use `tox`, which means we'd also need to figure out how to get it installed on GitHub Actions runners.
3. Try and see if we can get FabSim3 to use a standard approach for installation as a package so we could directly specific it as dependency as normal for the plugin and have it automatically installed in the isolated environment.

Of these, (3) is I'd say the 'nicest' solution in terms of adhering to good practice but I suspect will be non-trivial to get done. I think for now (1) is probably the best option as it should hopefully be relatively simple to implement and I think is less brittle than (2).

dleggat:
Agreed that (3) sounds ideal, but less immediately practical. I was trying to get something to work with (2), but couldn't work out how to get the right paths in the right places, and like you say it would have ended up very brittle either way. So I've put in a suppression for ImportError, that hopefully won't cause too many problems down the road?

FabNESO/__init__.py
```diff
@@ -1,5 +1,8 @@
 """Neptune Exploratory SOftware (NESO) plugin for FabSim3."""
+import contextlib
+
+with contextlib.suppress(ImportError):
+    from .tasks import neso, neso_ensemble
 
-from .tasks import neso, neso_ensemble
 
 __all__ = ["neso", "neso_ensemble"]
```
matt-graham: I would suggest outputting a warning if we cannot import from `.tasks` and only defining `__all__` attribute in this case, with something like

```Python
try:
    from .tasks import neso, neso_ensemble

    __all__ = ["neso", "neso_ensemble"]

except ImportError:
    import warnings

    warnings.warn(
        "Cannot import FabNESO tasks - fabsim.base.fab module may not be importable",
        stacklevel=2,
    )
```

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+    temp_conditions_path = Path(tmpdir) / "conditions.xml"
+    shutil.copyfile(
+        Path("config_files") / "two_stream" / "conditions.xml", temp_conditions_path
+    )
+
+    edit_parameters(temp_conditions_path, {parameter_to_test: parameter_test_value})
+
+    assert _check_parameter_in_conditions(
+        temp_conditions_path, parameter_to_test, parameter_test_value
+    )
```
matt-graham:
```suggestion
def test_edit_parameters(tmp_path: Path) -> None:
    """Test the edit_parameters method of ensemble_tools."""
    parameter_to_test = "particle_initial_velocity"
    parameter_test_value = 10.0
    temp_conditions_path = tmp_path / "conditions.xml"
    shutil.copyfile(
        Path(__file__).parents[1] / "config_files" / "two_stream" / "conditions.xml",
        temp_conditions_path,
    )
    edit_parameters(temp_conditions_path, {parameter_to_test: parameter_test_value})
    assert _check_parameter_in_conditions(
        temp_conditions_path, parameter_to_test, parameter_test_value
    )
```

Apologies my fault for suggesting using `tmpdir` without reading the Pytest documentation closely. It looks like [the `tmp_path` fixture is the currently recommended approach for creating temporary directories](https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmp-path-fixture), with `tmp_path` being a standard `pathlib.Path` object, avoiding the need to import `py.path.local` (with `py` package now considered legacy / in maintenance mode only).

We should also avoid assuming the tests are run from the root of the repository directory by either specifying the path to the `config_files` directory relative to the test module as here or by importing `FabNESO` top-level package and specifying relative to package as `Path(FabNESO.__file__).parents[1] / "config_files"`.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+    data = ElementTree.parse(conditions_file_name)
+    root = data.getroot()
+    conditions = root.find("CONDITIONS")
+    if conditions is None:
+        msg = f"Failed to find CONDITIONS in the file {conditions_file_name}"
+        raise ValueError(msg)
+    parameters = conditions.find("PARAMETERS")
+    if parameters is None:
+        msg = f"Failed to find PARAMETERS in the \
+        CONDITIONS node of {conditions_file_name}"
```
matt-graham:
```suggestion
        msg = (
            f"Failed to find PARAMETERS in the CONDITIONS node of "
            f"{conditions_file_name}"
        )
```
[PEP 8 guidance for wrapping long lines is](https://peps.python.org/pep-0008/#maximum-line-length)

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+        match = re.match(
+            r"\s*(?P<key>\w+)\s*=\s*(?P<value>-?\d*(\.?\d)+)\s*", element.text
+        )
+        if match is None:
+            msg = f"Parameter definition of unexpected format: {element.text}"
+            raise ValueError(msg)
+        key = match.group("key")
+        value = float(match.group("value"))
+        if key == parameter_name:
+            return expected_value == pytest.approx(value)  # Because of rounding errors
```
matt-graham: Returning as soon as a match is found checks that the specified parameter is defined in the conditions file but doesn't preclude it being defined multiple times with different values, which we possibly also want to test for (I think the behaviour of `edit_parameters` would currently be to set all instances equal to the overriden value).

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # If we make it this far then we haven't found the parameter we were looking for,
+    # which is surely a problem in its own right
+    msg = f"Could not find the parameter {parameter_name} in {conditions_file_name}"
+    raise ValueError(msg)
```
matt-graham: I would say better to return `False` in the case that parameter not found at all in file, and then have the `assert` statements in test handle this as a test failure. Generally tests are expected to fail by asserting something false rather than raise exceptions (which will be shown as a test error rather than exception).

pyproject.toml
```diff
@@ -19,6 +19,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
+    "defusedxml",
```
matt-graham: To avoid Ruff complaining about using `xml.ElementTree` we could add local `# noqa: S314` comments. Technically if `defusedxml` is only required in the tests it should be a optional development dependency rather than a direct dependency of package with something like
```TOML
optional-dependencies = {dev = ["defusedxml"]}
```
While we might want to just switch to using `defusedxml` in the package as well, for now I would just stick with standard library `xml` with local ignore comments and we look at switching to `defusedxml` in a separate PR. Because of the way FabSim3 handles plugins we will probably need to specify dependencies in a `requirements.txt` file and [then specify the `dependencies` property to be dynamic in the `pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata), which is all a bit tangential to adding the tests so I don't think worth doing in this PR.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+import itertools
+import re
+import shutil
+from pathlib import Path
+
+import pytest
+from defusedxml import ElementTree
+from py.path import local
+
+# from plugins.FabNESO.utils.make_sweep_dir
```
matt-graham:
```suggestion
```

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+        edit_file=edit_file,
+        parameter_to_scan=parameter_to_scan,
+        scan_range=scan_range,
+        outdir_prefix=outdir_prefix,
+    )
+
+    assert (test_sweep_dir / "SWEEP").is_dir()
+    assert len(list((test_sweep_dir / "SWEEP").iterdir())) == n_dirs
+    for i in range(n_dirs):
+        cond_file = test_sweep_dir / "SWEEP" / f"asd{i}" / "conditions.xml"
```
matt-graham:
```suggestion
        cond_file = test_sweep_dir / "SWEEP" / f"{outdir_prefix}{i}" / "conditions.xml"
```

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+        create_dir_tree(
+            sweep_path=test_sweep_dir,
+            n_dirs=n_dirs,
+            destructive=destructive,
+            copy_dir=copy_dir,
+            edit_file=edit_file,
+            parameter_to_scan=parameter_to_scan,
+            scan_range=scan_range,
+            outdir_prefix=outdir_prefix,
+        )
```
matt-graham: I would say this is logically distinct from the previous part of the test so should be in a separate test. Good idea to test exception behaviour though.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+    assert (test_sweep_dir / "SWEEP").is_dir()
+    assert len(list((test_sweep_dir / "SWEEP").iterdir())) == n_dirs
+    for i in range(n_dirs):
+        cond_file = test_sweep_dir / "SWEEP" / f"asd{i}" / "conditions.xml"
+        assert cond_file.is_file()
+        para_value = scan_range[0] + (i / (n_dirs - 1)) * (
+            scan_range[1] - scan_range[0]
+        )
```
matt-graham: Having to repeat the logic for computing the parameter value in the scan range here in the test probably suggest we should factor this out in to a separate function in `ensemble_tools` module, have a unit test for that function (checking for example it gives parameter values within scan range, with equal spacings) and then use that function both in the test and implementation to avoid the duplication. Not necessary to do for now but something for us to consider in future.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+def test_create_dir_tree(tmpdir: local) -> None:
+    """Test the create_dir_tree method of ensemble_tools."""
+    test_sweep_dir = Path(tmpdir) / "test_sweep"
+    n_dirs = 5
+    destructive = True
+    copy_dir = Path("config_files") / "two_stream"
+    edit_file = "conditions.xml"
+    parameter_to_scan = "particle_initial_velocity"
+    scan_range = (0.2, 5.0)
+    outdir_prefix = "asd"
```
matt-graham: Ideally I would make at least some of these test arguments and use `pytest.mark.parameterize` to test over multiple values - probably `destructive` (to make sure both branches of implementation are tested), `n_dirs` (to make sure this works for edge case like `ndirs=1` as well as more typical values) and `copy_dir` / `parameter_to_scan` (which probably will need to be jointly parameterized) to ensure this is robust to different base conditions files.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+        if key == parameter_name:
+            return expected_value == pytest.approx(value)  # Because of rounding errors
+
+    # If we make it this far then we haven't found the parameter we were looking for,
+    # which is surely a problem in its own right
+    msg = f"Could not find the parameter {parameter_name} in {conditions_file_name}"
+    raise ValueError(msg)
+
+
+def test_create_dir_tree(tmpdir: local) -> None:
```
matt-graham: As above this should ideally be refactored to use `temp_path` - sorry again for the false lead on this.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+    n_total_directories = n_divs ** len(parameter_dict)
+    # Check we make the corect number of directories
+    assert len(list((sweep_path / "SWEEP").iterdir())) == n_total_directories
+
+    # Loop through the directories and check the conditions file
+    for indices in itertools.product(*(range(n_divs),) * len(parameter_dict)):
+        directory_name = "-".join(
+            f"{k}_{i}" for k, i in zip(parameter_dict, indices, strict=True)
+        )
```
matt-graham, line 235: As above, the repetition of code from the implementation here suggests we should be factoring this out in to a separate function with its own unit test and using across both the test and implementation. In this case we probably would want the unit test to check that a function for generating directory name generates unique names for each of a set of indices, and potentially includes the varied parameter names, but not for the specific format (as then if we decide to update how the names are created we have to separately update both the test and implementation).

matt-graham: Again probably not worth changing `ensemble tools` here but we should look at doing some refactoring in a separate PR.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+                scan_range[1] - scan_range[0]
+            )
+            assert _check_parameter_in_conditions(
+                this_dir / "conditions.xml", parameter, para_value
+            )
+            with pytest.raises(ValueError, match=r".*CONDITIONS.*"):
+                # The mesh file should be not edited
+                assert _check_parameter_in_conditions(
+                    this_dir / "mesh.xml", parameter, para_value
+                )
```
matt-graham: Not sure if it makes sense to have an `assert` within the `pytest.raises` here, and again I would probably separate out checks for expected exception behaviour  in to a separate test function.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+            f"{k}_{i}" for k, i in zip(parameter_dict, indices, strict=True)
+        )
+        this_dir = sweep_path / "SWEEP" / directory_name
+
+        # Check the directory exists
+        assert this_dir.is_dir()
+
+        # Check the conditions and mesh files exist
+        assert (this_dir / "conditions.xml").is_file()
+        assert (this_dir / "mesh.xml").is_file()
```
matt-graham: Rather than checking specifically for presence of `conditions.xml` and `mesh.xml`, we could avoid hardcoding the assumption of these filenames in this part of the test by checking that the filenames inf `copy_dir` and `this_dir` are the same.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+            destructive=destructive,
+            copy_dir=copy_dir,
+            edit_file=edit_file,
+            parameter_to_scan=parameter_to_scan,
+            scan_range=scan_range,
+            outdir_prefix=outdir_prefix,
+        )
+
+
+def test_create_dict_sweep(tmpdir: local) -> None:
```
matt-graham: Again replace `tmpdir` with `temp_path`

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,163 @@
...
+    """Test the create_dict_sweep method of ensemble_tools."""
+    sweep_path = Path(tmpdir) / "test"
+    n_divs = 5
+    destructive = True
+    copy_dir = Path("config_files") / "two_stream"
+    edit_file = "conditions.xml"
+    parameter_dict = {
+        "particle_initial_velocity": (0.1, 2.5),
+        "particle_charge_density": (102.0, 108.0),
+    }
```
matt-graham: As in earlier comment for `test_create_dir_tree` I would parameterize `n_divs` here to ensure we test edge case `n_divs = 1` as well as typical case, and probably `copy_dir` / `parameter_dict` and `destructive`.

matt-graham (changes requested):
Thanks @dleggat this is looking good - I've made some suggestions about separating out some of the checks in to different test functions (generally tests which check one or a few related conditions rather than lots of logically distinct conditions are preferred from a perspective of failure in test reports then being more granular and helping to identify the specific issue needing to be addressed) and also in parameterizing some tests to cover a wide range of cases.

tests/test_ensemble_tools.py
dleggat: What I'm thinking instead then (to also address the next comment) is to return a `tuple` of `(n_equal, n_different)` where `n_equal` is the number of times we found the parameter with the expected value, and `n_different` is the number of times the parameter appears with a different value.

Then we can also use the test to make sure the parameter is found and that there are no additional or incorrect matches.

tests/test_ensemble_tools.py
dleggat: See previous discussion. This would then return `(0,0)`, which can be a failed `assert` in a subsequent test.

pyproject.toml
dleggat: Have switched back to `xml` for now.

tests/test_ensemble_tools.py
dleggat: So I put this in, and discovered that there is, indeed, a `ZeroDivisionError` for `n_dirs = 1` in the `ensemble_tools`. So I've included a hotfix for that in this PR.

dleggat: I have parametrized several of these, let me know if there are further test cases you'd like considered

tests/test_ensemble_tools.py
dleggat: I've added a `test_exceptions_create_dir_tree` method to address a number of possible exceptions and test that we're treating them correctly instead. As the create_dict_sweep method doesn't throw any exceptions of its own, I've just removed the one I had there and not included any further checks for that method.

tests/test_ensemble_tools.py
dleggat: Similarly, I've parametrized those mentioned variables, I can add further test cases if there are any other obvious edge cases you can think of.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,239 @@
...
+    shutil.copyfile(
+        Path(__file__).parents[1] / "config_files" / "two_stream" / "conditions.xml",
+        temp_conditions_path,
+    )
+
+    edit_parameters(temp_conditions_path, {parameter_to_test: parameter_test_value})
+
+    assert _check_parameter_in_conditions(
+        temp_conditions_path, parameter_to_test, parameter_test_value
+    ) == (1, 0)
```
matt-graham:
```suggestion
    n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
        temp_conditions_path, parameter_to_test, parameter_test_value
    )
    assert n_equal_in_value == 1 and n_different_in_value == 0
```
Good idea to have the check function output two different values here. To help make the test a bit more self documenting in case of a failure (as Pytest will show this assert statement on a failure) it might be good to attach names to the outputs which make it clearer why we expect values of 1 and 0.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,239 @@
...
+    create_dir_tree(
+        sweep_path=input_dict["sweep_path"],
+        n_dirs=input_dict["n_dirs"],
+        destructive=input_dict["destructive"],
+        copy_dir=input_dict["copy_dir"],
+        edit_file=input_dict["edit_file"],
+        parameter_to_scan=input_dict["parameter_to_scan"],
+        scan_range=input_dict["scan_range"],
+        outdir_prefix=input_dict["outdir_prefix"],
+    )
```
matt-graham: Rather than have this as a separate function you could use [argument unpacking](https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists) with the dictionary to get the same effect - `create_dir_tree(**input_dict)`.

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,239 @@
...
+            cond_file, parameter_to_scan, para_value
+        ) == (1, 0)
+        # Use an incorrect value of the parameter to get (0,1)
+        assert _check_parameter_in_conditions(
+            cond_file, parameter_to_scan, para_value + 10.0
+        ) == (0, 1)
+        # Make up a parameter and check we get (0,0)
+        assert _check_parameter_in_conditions(
+            cond_file, "fake_parameter", para_value
+        ) == (0, 0)
```
matt-graham: Again would either explicitly attach meaningful names to the outputs of `_check_parameter_in_conditions` calls here, or make it explicit in comments - for example
```Python
# Check the parameter has been edited correctly and appears only once
# that is (n_equal_in_value, n_different_in_value) = (1, 0)
```

tests/test_ensemble_tools.py
```diff
@@ -0,0 +1,239 @@
...
+            scan_range = parameter_dict[parameter]
+            para_value = (
+                scan_range[0]
+                if n_divs == 1
+                else scan_range[0]
+                + (indices[i] / (n_divs - 1)) * (scan_range[1] - scan_range[0])
+            )
+            assert _check_parameter_in_conditions(
+                this_dir / "conditions.xml", parameter, para_value
+            ) == (1, 0)
```
matt-graham:
```suggestion
            n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
                this_dir / "conditions.xml", parameter, para_value
            )
            assert n_equal_in_value == 1 and n_different_in_value == 0
```

tests/test_ensemble_tools.py
matt-graham:
```suggestion
        # Check the parameter has been edited correctly and appears only once
        n_equal_in_value, n_different_in_value =  _check_parameter_in_conditions(
            cond_file, parameter_to_scan, para_value
        ) 
        assert n_equal_in_value == 1 and n_different_in_value == 0
```

Thinking about this a bit more, the latter two tests here are more of a test of the `_check_parameter_in_conditions` function itself (which is a great idea to have) as they're checking it gives the expected output in situations outside the core test condition. I would therefore remove them from this specific test and factor them out in to a separate test to make it clearer if there is a failure here, it is likely that the thing needing fixing is the check function - something like the following

```Python
def test_check_parameter_in_conditions(tmp_path: Path) -> None:
    """Test function for checking parameter definitions in conditions file."""
    parameter_to_test = "particle_initial_velocity"
    parameter_test_value = 10.0
    temp_conditions_path = tmp_path / "conditions.xml"
    shutil.copyfile(
        Path(__file__).parents[1] / "config_files" / "two_stream" / "conditions.xml",
        temp_conditions_path,
    )
    edit_parameters(temp_conditions_path, {parameter_to_test: parameter_test_value})
    # Use different value of the parameter and check counted correctly
    n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
        cond_file, parameter_to_test, parameter_test_value + 10.0
    )
    assert n_equal_in_value == 0 and n_different_in_value == 1
    # Make up a parameter and check we get zero counts
    n_equal_in_value, n_different_in_value = _check_parameter_in_conditions(
        cond_file, "fake_parameter", parameter_test_value 
    )
    assert n_equal_in_value == 0 and n_different_in_value == 0
```

tests/test_ensemble_tools.py
matt-graham: Looks good thanks!

matt-graham (approved):
Thanks @dleggat for the updates. I've made a few small further suggestions but none are essential. I'll open issues for the points I made about factoring out some functions and adding unit tests for these.

tests/test_ensemble_tools.py
dleggat: So this is exactly what I was trying to originally - each of the `_call_create_dir_tree_from_dict(argument_dict)` calls were `create_dict_tree(**argument_dict)` - but `mypy` complained a lot about the type hinting for `create_dir_tree` not properly matching the dict arguments passed. I got the following errors for each instance:

```
tests/test_ensemble_tools.py:158: error: Argument 1 to "create_dir_tree" has incompatible type "**dict[str, object]"; expected "Path"  [arg-type]
tests/test_ensemble_tools.py:158: error: Argument 1 to "create_dir_tree" has incompatible type "**dict[str, object]"; expected "int"  [arg-type]
tests/test_ensemble_tools.py:158: error: Argument 1 to "create_dir_tree" has incompatible type "**dict[str, object]"; expected "bool"  [arg-type]
tests/test_ensemble_tools.py:158: error: Argument 1 to "create_dir_tree" has incompatible type "**dict[str, object]"; expected "str"  [arg-type]
tests/test_ensemble_tools.py:158: error: Argument 1 to "create_dir_tree" has incompatible type "**dict[str, object]"; expected "tuple[float, float]"  [arg-type]
Found 5 errors in 1 file (checked 1 source file)

``` 

I tried explicitly casting the objects for each keyword in `argument_dict`, but it didn't help. To avoid writing out the entire call each time I then refactored into the `_call_create_dir_tree_from_dict` method. If you know a way to avoid these errors in `mypy` I'd be happy to put it back!

matt-graham: Ah I'd forgotten `mypy` is strict about this. One option would be to create a `TypedDict` instance, something like
```Python
class _CreateDirTreeArgsDict(TypedDict):
    sweep_path: Path
    n_dirs: int
    destructive: bool
    copy_dir: Path
    edit_file: str
    parameter_to_scan: str
    scan_range: tuple[float, float]
    outdir_prefix: str
```
and then swap `argument_dict` definition to
```Python
    argument_dict = _CreateDirTreeArgsDict(
        sweep_path=tmp_path / "test_sweep",
        n_dirs=5,
        destructive=True,
        copy_dir=Path("config_files") / "two_stream",
        edit_file="conditions.xml",
        parameter_to_scan="particle_initial_velocity",
        scan_range=(0.2, 5.0),
        outdir_prefix="asd",
    )
```
which should stop `mypy` complaining when unpacking. I think at the moment this would still leave an error when the `parameter_to_scan` argument is set to `None` as the type annotation for that argument says it should be a string rather than `None` type or a `str` (either wrap in `Optional` or `Union[..., None]`).

dleggat: Yeah, `mypy` doesn't like `parameter_to_scan` being set to `None`. It also doesn't like the type hint in `_CreateDirTreeArgsDict` being ` str | None ` because it is strictly `str` in `create_dir_tree`.

Since it shouldn't really be possible to get a `None` value now, I've added in an extra exception to the `ensemble_tools` such that when `parameter_to_scan` value is left blank (the default action in the `make_sweep_dir` argument parser) it raises a ValueError, and changed the test to look for this. 

On a related note, I've also changed the `main` in `make_sweep_dir.py` so that if no `parameter_dict` is provided and no `para_to_template` is set, it will raise an exception instead of blindly running the `create_dir_tree` method.

matt-graham: Thanks, that all sounds sensible

tests/test_ensemble_tools.py
```diff
@@ -69,6 +70,30 @@ def _check_parameter_in_conditions(
     return (n_equals, n_different)
 
 
+def test_check_parameter_in_conditions(tmp_path: Path) -> None:
+    """Test that the private xml parser method of this class."""
```
matt-graham: This docstring doesn't read quite right - possibly an additional `that`?

dleggat: Yep, I think I changed my mind on what it was gonna say half way through writing it - woops! Fixed now

PR #28, opened 2023-12-07
Refactoring ensemble tools

### LINKED ISSUE #25: Factor out logic for computing parameter grids and add test
Having to repeat the logic for computing the parameter value in the scan range here in the test probably suggest we should factor this out in to a separate function in `ensemble_tools` module, have a unit test for that function (checking for example it gives parameter values within scan range, with equal spacings) and then use that function both in the test and implementation to avoid the duplication. Not necessary to do for now but something for us to consider in future.

_Originally posted by @matt-graham in https://github.com/UCL/FabNESO/pull/23#discussion_r1410929185_

### LINKED ISSUE #26: Factor out logic for setting sweep directory names and add test
As above, the repetition of code from the implementation here suggests we should be factoring this out in to a separate function with its own unit test and using across both the test and implementation. In this case we probably would want the unit test to check that a function for generating directory name generates unique names for each of a set of indices, and potentially includes the varied parameter names, but not for the specific format (as then if we decide to update how the names are created we have to separately update both the test and implementation).

_Originally posted by @matt-graham in https://github.com/UCL/FabNESO/pull/23#discussion_r1410943651_

dleggat:
Resolves #25 and #26 , refactoring the calculation of parameter grid values and directory names, and including unit tests for these new methods.

Also refactors out the xml parser from the tests, so that this method can be used later in the VBMC code.

Regarding the refactoring of calculate_parameter_value (issue #25); I considered making the method return an iterator, but the logic for the dictionary sweep methods didn't seem compatible with this, so I left it instead as a direct calculation of the parameter value at a point instead. Open to alternative suggestions on that, though.

### DIFF
--- a/FabNESO/ensemble_tools.py
+++ b/FabNESO/ensemble_tools.py
@@ -107,15 +107,15 @@ def create_dict_sweep(
         indices_iterator(n_dirs, len(parameter_grids)),
         strict=True,
     ):
-        directory_name = return_directory_name(parameter_values, indices)
+        directory_name = return_directory_name(list(parameter_values.keys()), indices)
         directory_path = Path(sweep_path) / "SWEEP" / directory_name
         shutil.copytree(copy_dir, directory_path)
         edit_parameters(directory_path / edit_file, parameter_values)
 
 
-def return_directory_name(parameter_values: dict, indices: tuple[int, ...]) -> str:
+def return_directory_name(parameter_names: list[str], indices: tuple[int, ...]) -> str:
     """Return the directory name given parameter names and indices."""
-    return "-".join(f"{k}_{i}" for k, i in zip(parameter_values, indices, strict=True))
+    return "-".join(f"{k}_{i}" for k, i in zip(parameter_names, indices, strict=True))
 
 
 def list_parameter_values(conditions_file: Path, parameter_name: str) -> list[str]:
--- a/FabNESO/ensemble_tools.py
+++ b/FabNESO/ensemble_tools.py
@@ -78,6 +78,11 @@ def _product_dict(input_dict: dict) -> Iterator[dict]:
         yield dict(zip(keys, values, strict=True))
 
 
+def indices_iterator(n_dirs: int, n_parameters: int) -> Iterator[tuple[Any, ...]]:
+    """Create an iterator for the indices of the dictionary sweep."""
+    yield from itertools.product(*(range(n_dirs),) * n_parameters)
+
+
 def create_dict_sweep(
     *,
     sweep_path: Path,
@@ -99,7 +104,7 @@ def create_dict_sweep(
     # Compute Cartesian products of all parameter value combinations plus grid indices
     for parameter_values, indices in zip(
         _product_dict(parameter_grids),
-        itertools.product(*(range(n_dirs),) * len(parameter_dict)),
+        indices_iterator(n_dirs, len(parameter_grids)),
         strict=True,
     ):
         directory_name = return_directory_name(parameter_values, indices)
--- a/FabNESO/ensemble_tools.py
+++ b/FabNESO/ensemble_tools.py
@@ -48,19 +48,27 @@ def create_dir_tree(
             raise FileExistsError(msg)
     sweep_path.mkdir(parents=True)
 
-    # Set the initial value of the scanned parameter to the lower limit
-    para_val = scan_range[0]
-
     for i in range(n_dirs):
+        para_val = calculate_parameter_value(n_dirs, scan_range[0], scan_range[1], i)
         new_dir = Path(sweep_path) / "SWEEP" / f"{outdir_prefix}{i}"
         shutil.copytree(copy_dir, new_dir)
         # Now we edit the parameter file for our
         # template scan if we're doing that
         edit_parameters(new_dir / edit_file, {parameter_to_scan: para_val})
-        # iterate para_val
-        para_val += (
-            0 if n_dirs == 1 else (scan_range[1] - scan_range[0]) / float(n_dirs - 1)
-        )
+
+
+def calculate_parameter_value(
+    n_dirs: int,
+    initial_value: float,
+    final_value: float,
+    iteration: int,
+) -> float:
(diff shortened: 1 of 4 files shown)


dleggat:
Resolves #26

dleggat:
I've also gone ahead and refactored the indices calculation to a generator in ensemble_tools, since we were using that in several different places too.

FabNESO/ensemble_tools.py
```diff
@@ -70,10 +78,15 @@ def _product_dict(input_dict: dict) -> Iterator[dict]:
         yield dict(zip(keys, values, strict=True))
 
 
+def indices_iterator(n_dirs: int, n_parameters: int) -> Iterator[tuple[Any, ...]]:
```
matt-graham:
```suggestion
def indices_iterator(n_dirs: int, n_parameters: int) -> Iterator[tuple[int, ...]]:
```

I think the tuples iterated over will always contain integers so type hint can be more specific here? Wasn't sure if `mypy` would complain about this if type hint for `itertools.product` is too broad, but from testing locally seems to work okay. Import of `Any` can also be removed in this case (along with adjustment to other `Any` usages below).

FabNESO/ensemble_tools.py
```diff
@@ -84,28 +97,58 @@ def create_dict_sweep(
...
-        directory_name = "-".join(
-            f"{k}_{i}" for k, i in zip(parameter_values, indices, strict=True)
-        )
+        directory_name = return_directory_name(parameter_values, indices)
         directory_path = Path(sweep_path) / "SWEEP" / directory_name
         shutil.copytree(copy_dir, directory_path)
         edit_parameters(directory_path / edit_file, parameter_values)
 
 
+def return_directory_name(parameter_values: dict, indices: tuple[Any, ...]) -> str:
```
matt-graham:
```suggestion
def return_directory_name(parameter_values: dict, indices: tuple[int, ...]) -> str:
```

tests/test_ensemble_tools.py
```diff
@@ -246,14 +247,77 @@ def test_create_dict_sweep(
...
+    assert len(directory_names) == n_dirs ** len(parameter_list)
+    # Check that we've made unique directories
+    n_unique_dirs = len(set(directory_names))
+    assert len(directory_names) == n_unique_dirs
+
+
+@pytest.mark.parametrize("n_dirs", [1, 3, 7])
+@pytest.mark.parametrize("n_parameters", [1, 2, 5, 7])
+def test_indices_iterator(n_dirs: int, n_parameters: int) -> None:
+    """The the indices_iterator from the ensemble_tools."""
```
matt-graham:
```suggestion
    """Test the indices_iterator from the ensemble_tools."""
```

tests/test_ensemble_tools.py
```diff
@@ -246,14 +247,77 @@ def test_create_dict_sweep(
...
+        {"particle_initial_velocity": [], "particle_charge_density": []},
+        {
+            "particle_initial_velocity": [],
+            "particle_charge_density": [],
+            "particle_number_density": [],
+        },
+    ],
+)
+def test_return_directory_name(n_dirs: int, parameter_list: dict) -> None:
+    """Test the return_directory_name method."""
```
matt-graham:
```suggestion
    """Test the return_directory_name function."""
```

FabNESO/ensemble_tools.py
```diff
@@ -84,28 +97,58 @@ def create_dict_sweep(
...
+    """Return the directory name given parameter names and indices."""
+    return "-".join(f"{k}_{i}" for k, i in zip(parameter_values, indices, strict=True))
```
matt-graham: As this function only requires the keys (parameter names) from `parameter_values` dictionary, I think it would be more readable and also simplify testing to just accept an iterable of strings as an argument `parameter_names`, and when calling just pass `parameter_values.keys()`.

FabNESO/make_sweep_dir.py
```diff
@@ -73,7 +73,7 @@ def main() -> None:
         # Use the dict to create a sweep directory
         create_dict_sweep(
             sweep_path=args.sweep_path,
-            n_divs=args.n_divs,
+            n_dirs=args.n_divs,
```
matt-graham: For consistency it might be good to change the name of the command line argument as well to `n_divs` to avoid users getting mixed up between the function and command line interfaces.

tests/test_ensemble_tools.py
```diff
@@ -246,14 +247,77 @@ def test_create_dict_sweep(
...
+    "parameter_list",
+    [
+        {"particle_initial_velocity": [], "particle_charge_density": []},
+        {
+            "particle_initial_velocity": [],
+            "particle_charge_density": [],
+            "particle_number_density": [],
+        },
+    ],
+)
```
matt-graham: If `return_directory_name` is changed to accept an iterable of names instead of a dictionary then the dummy values could be removed here and just lists / tuples of parameter names used.

FabNESO/ensemble_tools.py
matt-graham: This shouldn't require any changes to implementation other than changing parameter name and updating type hint.

matt-graham (changes requested):
Thanks @dleggat, this looks great. Have made a few minor comments and suggestions.

FabNESO/ensemble_tools.py
dleggat: Sure, `mypy` seems to be happy with this, so I've done as you suggest.

FabNESO/ensemble_tools.py
dleggat: I have had to cast the `.keys()` as a list to satiate `mypy`, but otherwise this has been done.

matt-graham: That works - in case it's helpful in future `Iterable` in `collections.abc` should I think be a valid type hint for either a list or `dict_keys` (or any object which implements an `__iter__` method).

matt-graham (approved):

PR #30, opened 2023-12-08
First draft of integrating PyVBMC into the FabNESO plugin

dleggat:
First draft of integrating `PyVBMC` into the plugin.

So far the instance runs locally correctly, though I suspect there are going to be problems with scheduler services that I will look to address before making this a full PR. Also everything is very hard coded at the moment, and I'll look into getting better abstraction for everything before the real PR. Just wanted to get the first draft up at least.

In addition, I've added in an option to the `edit_parameters` method of the `ensemble_tools` to allow it to create parameters that it didn't find in the first parse of changes. This was done in order to make sure that we're reading out the field density correctly, but may have further applications outside of this.


FabNESO/tasks.py
```diff
@@ -252,3 +267,199 @@ def neso_ensemble(
...
+    config_dict: dict[str, Any],
+) -> dict[str, np.ndarray]:
+    """Run a single instance of the NESO solver and return the observed_field."""
+    # If we're running remotely, tell the submission to wait until done
+    if "archer2" in fab.env.remote:
+        fab.update_environment(
+            {
+                "job_dispatch": "cd /work/$project/$project/$username ; sbatch --wait",
+            }
+        )
```
dleggat, line 520: This is how I'm currently forcing the code to wait after `sbatch` submission. `job_dispatch` is an environment variable that can be edited in the `machines_FabNESO_user.yml` file, and I toyed with just adding it to the example file there, but this feels like a more safe way to force the wait, given the current proposal that we just want it to work on Archer2 for now.

I also tested using the `wait_complete()` method built in to `FabSIM`, and that works perfectly fine (and would be platform independent), but it does have much larger overheard when running short jobs.

dleggat:
This version of the code runs the `neso_vbmc` task wherever `FabSIM` is being run, and submits each iteration of `NESO` to the designated scheduler (archer2 for now). Thanks to archer2's TOTP system we don't have to worry about signing in for each job submission, but the I would suggest running the task in a `screen` session to avoid accidental closing or timeouts of the terminal.

As far as I've been able to see, PyVBMC seems to run in a very serial manner, so even though it runs some number of NESO simulations per iteration (5 by default), I can't see a way of making those tasks run in parallel, and am currently just waiting for each of them to finish in sequence. This doesn't feel particularly optimal?

FabNESO/tasks.py
matt-graham: This seems a good solution for now. Agree that setting as part of task rather than having user need to override in the YAML file is preferable from a robustness perspective and in still allowing asynchronous dispatch of jobs on ARCHER2 in other tasks. Good to also know the `wait_complete` approach is viable - we could possible look at putting that as a fallback option when not using a Slurm based scheduler in future.

matt-graham:
> As far as I've been able to see, PyVBMC seems to run in a very serial manner, so even though it runs some number of NESO simulations per iteration (5 by default), I can't see a way of making those tasks run in parallel, and am currently just waiting for each of them to finish in sequence. This doesn't feel particularly optimal?

Yes this is a big limitation of PyVBMC's approach, and agree is suboptimal in our use case where we'd like to be able to submit batches of jobs to run in parallel. I did look previously to see if it would be possible to parallelise the evaluation of the model in PyVBMC, but as you say it's not obvious how to do this, as there are operations performed using model output after each of the evaluations in a particular iteration. Basically I think the issue is that inherently the criterion PyVBMC uses to choose the next inputs to evaluate at is sequential in nature.

With the in mind I have been working on writing [a Python implementation](https://github.com/UCL/calibr) of the algorithm described in  _Parallel Gaussian Process Surrogate Bayesian Inference with Noisy Likelihood Evaluations_ (Järvenpää, Gutmann, Vehtari and Marttinen) [📄 paper](https://doi.org/10.1214/20-BA1200), [💻 code](https://github.com/mjarvenpaa/parallel-GP-SL). The algorithm in that paper is specifically designed to choose batches of inputs to evaluate the model at in parallel, and uses a similar approach overall to PyVBMC (inasmuch as iteratively fits a Gaussian process emulator to the target posterior distribution). Annoyingly the reference MATLAB code doesn't actually allow parallel evaluation of the model, and also being in MATLAB is a bit of pain to get things set up. I'm writing the Python implementation using JAX which means we can get derivatives of the various functions we need to optimize / perform MCMC with for free (and potentially accelerate on a GPU). I've got as far getting the overall fitting loop to work with a toy 2D Rosenbrock function like model :tada: in a notebook, but need to start abstracting this out in reusable library functions. I'm aiming to make the actual interface similar to PyVBMC, beyond allowing to pass in a `async` function for simulating the model (which will mean we can set off batches of runs asynchronously on a scheduler), so hopefully we can use all you've done here to come up with something similar for this package.

FabNESO/read_outputs.py
```diff
@@ -0,0 +1,21 @@
+"""A script to read the outputs of the NESO solvers."""
```
matt-graham:
```suggestion
"""Functions for reading the outputs of the NESO solvers."""
```

As far as I can tell this is intended for use as a library module not as a script?

FabNESO/tasks.py
```diff
@@ -21,10 +27,14 @@
     from types import SimpleNamespace
 
     fab = SimpleNamespace(task=lambda f: f, load_plugin_env_vars=lambda _: lambda f: f)
+    template = SimpleNamespace(
+        task=lambda f: f, load_plugin_env_vars=lambda _: lambda f: f
+    )
```
matt-graham: Not sure this should be necessary as it looks like `template` is only used within function definitions? The slightly hacky approach I used for the `fab` module was only required there as we use the `task` and `load_plugin_env_vars` functions in the `fab` module as decorators, which will be evaluated at import time, so if these names are missing due to `fabsim` not being importable we will get a `NameError`. For `template` as this is I think not accessed at import time, there shouldn't be a need to create a placeholder (and if `fabsim` is not importable at run time the `_check_fab_module_imported` function should raise an error before we get to `template` call).

FabNESO/tasks.py
```diff
@@ -125,6 +137,7 @@ def neso(
             applicable when running on a multi-node system.
         wall_time: Maximum time to allow job to run for. Only applicable when submitting
             to a job scheduler.
+        create_missing_parameters: Force missing parameters in overrides to be added
```
```suggestion
        create_missing_parameters: Force parameters in ``parameter_overrides`` missing
            from conditions file to be added.
```
Just to make what 'missing' means here a bit more explicit

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+        nodes: Number of nodes to run on. Only applicable when running on a multi-node
+            system.
+        cpus_per_process: Number of processing units to use per process. Only
+            applicable when running on a multi-node system.
+        wall_time: Maximum time to allow job to run for. Only applicable when submitting
+            to a job scheduler.
+        **vbmc_parameters: The parameters to be scanned in the VBMC instance. The value
+            is a colon separated list of the lower and upper bounds of the parameter
+
+    """
```
```suggestion
    """
```suggestion
    Run variational Bayesian Monte Carlo (VBMC) to calibrate parameters of a NESO solver.
    
    The VBMC algorithm (Acerbi, 2018) is an approximate Bayesian inference method for
    efficient parameter calibration in expensive to simulate models. Here we use the Python
    implementation of the algorithm in the package PyVBMC (Huggins et al., 2023).

    Args:
        config: Directory with ensemble configuration information.
        solver: Which NESO solver to use.
        conditions_file_name: Name of conditions XML file in configuration directory.
        mesh_file_name: Name of mesh XML in configuration directory.
        processes: Number of processes to run in each job in the ensemble.
        nodes: Number of nodes to run on. Only applicable when running on a multi-node
            system.
        cpus_per_process: Number of processing units to use per process. Only
            applicable when running on a multi-node system.
        wall_time: Maximum time to allow job to run for. Only applicable when submitting
            to a job scheduler.
        **vbmc_parameters: The parameters to be calibrate using VBMC instance. Each
             value is a colon separated pair of lower and upper bounds, with the
             parameters assumed to be apriori uniformly distributed over this interval.

    References:
        1. Acerbi, L. (2018). Variational Bayesian Monte Carlo. 
           Advances in Neural Information Processing Systems, 31.
        2. Huggins et al., (2023). PyVBMC: Efficient Bayesian inference in Python.
           Journal of Open Source Software, 8(86), 5428, https://doi.org/10.21105/joss.05428
    """
```
Adding some extra detail

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+    config_dict = {
+        "config": config,
+        "para_overrides": {
+            # These ensure that the field information is written out, which is not done
+            # be default in the conditions files.
+            # This would need to be changed per solver
+            "particle_num_write_field_steps": 100,
+            "line_field_deriv_evaluations_step": 20,
+            "line_field_deriv_evaluations_numx": 100,
+            "line_field_deriv_evaluations_numy": 1,
```
matt-graham: We could potentially have a specific configuration directory for the purposes of calibration with these set there? This would make it easy for users to add their own configurations for use with different solvers. We could potentially also require that a Python script is defined there which computes the negative log density from the simulation outputs.

matt-graham: So in addition to `conditions_file_name` and `mesh_file_name` we would have something like `compute_log_target_density_script` which would be expected to be a path to a script which does something like take the directory the model output files are written to as an input argument and outputs to `stdout` the log density. Or alternatively we could make it an importable module which is required to define a function with a specified name and signature. Possibly not worth doing in this PR but something to consider for future.

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+
+    config_dict["parameters_to_scan"] = parameters_to_scan
+
+    # Set up the bounds information required by PyVBMC
+    bounds = list(zip(*parameters_to_scan.values(), strict=True))
+    lower_bounds = np.array(bounds[0])
+    upper_bounds = np.array(bounds[1])
+
+    plausible_lower_bounds = lower_bounds + (upper_bounds - lower_bounds) / 4
+    plausible_upper_bounds = upper_bounds - (upper_bounds - lower_bounds) / 4
```
matt-graham: In general the lower and upper bounds may be infinite (PVBMC allows this) and independent from the plausible bounds. I fixed the plausible bounds to the middle 50% of each interval previously just to try things out, but we'd probably ideally want to expose this as something the user can set. Again this probably is more suited to a subsequent PR.

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+
+    # Choose a random starting position
+    rng = np.random.default_rng()
+    theta_0 = rng.uniform(plausible_lower_bounds, plausible_upper_bounds)
+
+    # Run a nominal version of the model with central values to find
+    # the `true' field measurements. Ideally this would be measured data
+    config_dict["initial_run"] = run_instance_return_field(config_dict)
```
matt-graham: This is the main thing I think it would be worth avoiding hardcoding in the current PR. Ideally I would say we do something like store a file with observed outputs being calibrated against in the config directory and load this file as part of the task and pass the observed values array as an additional argument  to `log_density`, with we closing over this argument in the `lambda` expression. This would at least make it possible for the user to subsitute other measure or simulated data to calibrate against.

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+    config_dict["para_overrides"] = parameters
+
+    # Run an instance of NESO and calculate the measured field strength.
+    observed_results = run_instance_return_field(config_dict)
+
+    # Calculate the joint log likelihood using the reference field in the config_dict
+    return -(
+        (config_dict["initial_run"]["field_value"] - observed_results["field_value"])
+        ** 2
+        / (2 * config_dict["noise_factor"] ** 2)
```
```suggestion
        / (2 * config_dict["noise_factor"] ** 2)
```

I would say `observation_noise_std` is more explicit than `noise_factor`

FabNESO/tasks.py
```diff
@@ -252,3 +267,209 @@ def neso_ensemble(
...
+    theta: list[float],
+    config_dict: dict[str, Any],
+) -> list:
+    """Run an instance of the neso task and return the log_joint_density."""
+    parameters = dict(
+        zip(config_dict["parameters_to_scan"].keys(), theta, strict=True),
+        **config_dict["para_overrides"],
+    )
+
+    config_dict["para_overrides"] = parameters
```
matt-graham: Overwriting the previous value in `config_dict` could be a bit error prone as I think there will be a single `config_dict` instance which is shared across all calls to this function. The unpacking of `**config_dict["para_overrides"]` in the definition of `parameters` I think means any values written to `para_overrides` in a previous iteration might end up being used in place of the values extracted from `theta` (as `dict([("a", 1)], **{"a": 2})` evaluates to `{"a": 2}`). I would instead pass the `parameters` dict as a separate argument to `run_instance_return_field`  and use in place of the ` **config_dict["para_overrides"]` expression in the call to `neso`.

matt-graham (commented):
Thanks @dleggat, this is looking great. There are some things I've commented on that are more reminders for myself to create issues to work in future PR. If it's not too much work, it would be good if in this PR we could come up with an approach to avoid hardcoding the generation of the 'observed' data as part of the task by instead having the observations read from a file in the `config` directory (probably creating a specific configuration for calibration purposes), as this would feel more of a complete working example, even if in practice we're always going to be used simulated data atm.

FabNESO/tasks.py
dleggat: So I did consider this, and from what I could tell it was behaving correctly, but I think you're right in that we should treat it in a safer way. To that point I've added in the parameter overrides as an additional argument in `run_instance_return_field` as suggested.

FabNESO/tasks.py
dleggat: I have implemented this by reading in a numpy array from a reference file that acts as the reference during the calibration. In order to enable this, I've added in an extra task that lets the user run a single instance of NESO with parameters of their choosing, and saving the output to a specified file.

FabNESO/tasks.py
dleggat: I'm not sure if it did raise an error at some point (I'm not sure why I'd have thought to do the same thing otherwise...), but it doesn't seem to anymore, so I've taken out this part from the most recent version,

FabNESO/tasks.py
dleggat: It seemed straightforward to add the plausible upper and lower bounds to the colon separated list of parameters to scan, so I went ahead and added them in

FabNESO/tasks.py
```diff
@@ -252,3 +265,309 @@ def neso_ensemble(
...
+    config: str,
+    solver: str = "Electrostatic2D3V",
+    conditions_file_name: str = "conditions.xml",
+    mesh_file_name: str = "mesh.xml",
+    processes: int = 4,
+    nodes: int = 1,
+    cpus_per_process: int = 1,
+    wall_time: str = "00:15:00",
+    output_directory_name: str = "",
+    reference_field_file: str = "",
```
matt-graham, line 300: Rather than making this default to an empty string, I'd say probably better to not give a default here and have this as a required argument as that might suggest it doesn't need to be set.

FabNESO/tasks.py
```diff
@@ -252,3 +265,309 @@ def neso_ensemble(
...
+        conditions_file_name,
+        mesh_file_name,
+        processes,
+        nodes,
+        cpus_per_process,
+        wall_time,
+    )
+
+    # Make this read in from a config too?
+    observation_noise_std = 0.1
```
matt-graham, line 373: I think just having this passed in as an argument to the task would be the easiest way to let user change.

pyproject.toml
```diff
@@ -19,6 +19,8 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
+    "h5py",
+    "pyvbmc",
```
matt-graham, line 23: To get this to work nicely with FabSim3 `install_plugin` task we will need to pull these out in to a separate `requirements.txt` file and then have the `pyproject.toml` specify the `dependencies` property as dynamic and point to the file.

matt-graham (approved):
Thanks @dleggat for the updates. I've added a few extra comments of things that I should address at some point, but I think this is good to go in now as I've tested its working locally and I can address the remaining points in a subsequent PR(s).

matt-graham:
Just realised I've got the branch protection rules set up to not allow merging with unresolved conversation so I'll do what I should have done in the first place and make these in to issues 😁

Issue #16, opened 2023-11-27
Create HTML documentation on GitHub pages


matt-graham:
It would be useful to have basic autogenerated API documentation for the tasks and utility functions using something like Sphinx or Mkdocs, and GitHub Actions workflow for automatically building and deploying on pushes to `main` (and testing building on PRs).

PR #31, opened 2023-12-15
Extending the ensemble tools and task to allow variable parameter grids, and command line interaction

### LINKED ISSUE #20: Allow specifying different grid dimensions for each parameter
Current the `create_dict_sweep` function in `ensemble_tool` and by extension the `make_sweep_dir` script only allows specifying a single `n_divs` argument for setting the number of parameter values along each axis of the parameter grid. We might want to specify grids with variable resolutions across the different parameters. This should be easy to support by having `parameter_dict` argument values be tuples of three values, with the first value the start point of range, second value the end point and third value the number of linearly spaced points to use across range, equivalent to [the arguments to `numpy.linspace`](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html).

So for example `create_dict_sweep(..., parameter_dict={"particle_initial_velocity": (0.5, 1.5, 3), "particle_charge_density": (90, 110, 2)}")`

would correspond to the grid of parameter combinations
```
[
  {"particle_initial_velocity: 0.5, "particle_charge_density": 90},
  {"particle_initial_velocity: 0.5, "particle_charge_density": 100},
  {"particle_initial_velocity: 1.0, "particle_charge_density": 90},
  {"particle_initial_velocity: 1.0, "particle_charge_density": 100},
  {"particle_initial_velocity: 1.5, "particle_charge_density": 90},
  {"particle_initial_velocity: 1.5, "particle_charge_density": 100},
]
```

### LINKED ISSUE #21: Allow generating ensemble configurations directly from task
Currently the user is required to separately create the directory structure containing the 'sweep' / ensemble of configurations to run with the `neso_ensemble` task using the `make_sweep_dir` script. It would be nice to allow users to (perhaps optionally) directly specify parameter grids to run an ensemble over as part of the task arguments, for example assuming we implement something along the lines of what is suggested in #20 for specifying per-parameter grid dimensions, we could have a command like

```
fabsim neso_ensemble:two_stream,particle_initial_velocity=(0.5,1.5,3),particle_charge_density=(90,100,2)
```

which would run an ensemble over configurations based on the `two_stream` base configuration but with a grid of parameter overrides

```
[
  {"particle_initial_velocity: 0.5, "particle_charge_density": 90},
  {"particle_initial_velocity: 0.5, "particle_charge_density": 100},
  {"particle_initial_velocity: 1.0, "particle_charge_density": 90},
  {"particle_initial_velocity: 1.0, "particle_charge_density": 100},
  {"particle_initial_velocity: 1.5, "particle_charge_density": 90},
  {"particle_initial_velocity: 1.5, "particle_charge_density": 100},
]
```

To manage creating the ensemble directory dynamically we could use a similar temporary directory approach as used currently for the parameter overrides in the single instance `neso` task, with a temporary subdirectory in `configs` created and then `ensemble_tools.create_dict_sweep` used to populate this with the sweep directory using the specified base configuration and parameter grid spec.

dleggat:
Resolves #20, resolves #21 

`parameter_dict` in the `create_dict_sweep` function of `ensemble_tools` now includes an additional requirement of a number of steps for the sweep. In the `make_sweep_dir` script, this can either be set directly by the user when passing the input `dict` or assigned automatically with `n_dirs` if no argument is given.

This has necessitated a change to the way the iteration indices are calculated, which has been propagated throughout the `ensemble_tools` and its associated tests.

I have then also extended the `neso_ensemble` task to allow the definition of parameters to sweep directly in the command line, by calling this function in a manner similar to the original `neso` task.

A minor problem I encountered here was that even inside quotation and speech marks `FabSIM3` still splits the input argument list by commas. This means that when an argument such as `particle_initial_velocity="(0.2,2.0,5)"` is provided, `FabSIM` interprets this as `{"particle_initial_velocity": "\"(0.2"}` and discards the rest of the tuple. I couldn't find any obvious way around this, so I'm suggesting that for the command line definition of our sweep bounds the list be given with a `/` separating the parameters. This is then interpreted into the correct form inside the `neso_ensemble` task. The above example would then be: `particle_initial_velocity=0.2/2.0/5`. If that sounds acceptable, I can update the documentation to reflect this.

### DIFF
--- a/FabNESO/ensemble_tools.py
+++ b/FabNESO/ensemble_tools.py
@@ -104,10 +104,9 @@ def create_dict_sweep(
         for key, (low, high, n_dir_parameter) in parameter_dict.items()
     }
     # Compute Cartesian products of all parameter value combinations plus grid indices
-    [n_dir[2] for n_dir in parameter_dict.values()]
     for parameter_values, indices in zip(
         _product_dict(parameter_grids),
-        indices_iterator([n_dirs[2] for n_dirs in parameter_dict.values()]),
+        indices_iterator([n_dir for *_, n_dir in parameter_dict.values()]),
         strict=True,
     ):
         directory_name = return_directory_name(list(parameter_values.keys()), indices)
--- a/FabNESO/ensemble_tools.py
+++ b/FabNESO/ensemble_tools.py
@@ -78,19 +78,18 @@ def _product_dict(input_dict: dict) -> Iterator[dict]:
         yield dict(zip(keys, values, strict=True))
 
 
-def indices_iterator(n_dirs: int, n_parameters: int) -> Iterator[tuple[int, ...]]:
+def indices_iterator(n_dirs: list[int]) -> Iterator[tuple[int, ...]]:
     """Create an iterator for the indices of the dictionary sweep."""
-    yield from itertools.product(*(range(n_dirs),) * n_parameters)
+    yield from itertools.product(*[range(n_dir) for n_dir in n_dirs])
 
 
 def create_dict_sweep(
     *,
     sweep_path: Path,
-    n_dirs: int,
     destructive: bool,
     copy_dir: Path,
     edit_file: str,
-    parameter_dict: dict[str, tuple[float, float]],
+    parameter_dict: dict[str, tuple[float, float, int]],
 ) -> None:
     """Use a dictionary with each parameter interval to create a sweep directory."""
     # If destructive, delete the whole tree if it already exists
@@ -98,13 +97,17 @@ def create_dict_sweep(
         shutil.rmtree(sweep_path)
     # Uniformly spaced grids on [low, high] for each parameter
     parameter_grids = {
-        key: [calculate_parameter_value(n_dirs, low, high, i) for i in range(n_dirs)]
-        for key, (low, high) in parameter_dict.items()
+        key: [
+            calculate_parameter_value(n_dir_parameter, low, high, i)
+            for i in range(n_dir_parameter)
+        ]
+        for key, (low, high, n_dir_parameter) in parameter_dict.items()
     }
     # Compute Cartesian products of all parameter value combinations plus grid indices
+    [n_dir[2] for n_dir in parameter_dict.values()]
     for parameter_values, indices in zip(
         _product_dict(parameter_grids),
-        indices_iterator(n_dirs, len(parameter_grids)),
+        indices_iterator([n_dirs[2] for n_dirs in parameter_dict.values()]),
         strict=True,
     ):
         directory_name = return_directory_name(list(parameter_values.keys()), indices)
--- a/FabNESO/make_sweep_dir.py
+++ b/FabNESO/make_sweep_dir.py
@@ -4,7 +4,7 @@ import argparse
 from ast import literal_eval
 from pathlib import Path
 
-from .ensemble_tools import create_dict_sweep, create_dir_tree
+from FabNESO.ensemble_tools import create_dict_sweep, create_dir_tree
 
 
 def main() -> None:
@@ -70,10 +70,14 @@ def main() -> None:
         if not isinstance(parameter_dict, dict):
             msg = "Did not receive a dict as input for parameter_dict"
             raise ValueError(msg)
+        # If we define n_dirs in the command line argument, override any
+        # dict parameters we didn't assign n_dirs to
+        for _parameter, values in parameter_dict.items():
+            if len(values) < 3:  # noqa: PLR2004
+                values.append(args.n_dirs)
         # Use the dict to create a sweep directory
         create_dict_sweep(
             sweep_path=args.sweep_path,
-            n_dirs=args.n_dirs,
             destructive=args.destructive,
             copy_dir=args.copy_dir,
             edit_file=args.edit_file,
--- a/FabNESO/make_sweep_dir.py
+++ b/FabNESO/make_sweep_dir.py
@@ -4,7 +4,7 @@ import argparse
 from ast import literal_eval
 from pathlib import Path
 
-from FabNESO.ensemble_tools import create_dict_sweep, create_dir_tree
+from .ensemble_tools import create_dict_sweep, create_dir_tree
 
 
 def main() -> None:
@@ -72,7 +72,7 @@ def main() -> None:
             raise ValueError(msg)
         # If we define n_dirs in the command line argument, override any
         # dict parameters we didn't assign n_dirs to
-        for _parameter, values in parameter_dict.items():
+        for values in parameter_dict.values():
             if len(values) < 3:  # noqa: PLR2004
                 values.append(args.n_dirs)
         # Use the dict to create a sweep directory
--- a/FabNESO/tasks.py
+++ b/FabNESO/tasks.py
@@ -15,7 +15,7 @@ except ImportError:
     from base import fab
 
 
-from .ensemble_tools import edit_parameters
+from .ensemble_tools import create_dict_sweep, edit_parameters
 
 fab.add_local_paths("FabNESO")
 
@@ -81,6 +81,7 @@ def neso_ensemble(
     solver: str = "Electrostatic2D3V",
     conditions_file_name: str = "conditions.xml",
     mesh_file_name: str = "mesh.xml",
+    **parameter_scans: str,
 ) -> None:
     """
     Run ensemble of NESO solver instances.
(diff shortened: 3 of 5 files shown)


matt-graham:
> A minor problem I encountered here was that even inside quotation and speech marks `FabSIM3` still splits the input argument list by commas. This means that when an argument such as `particle_initial_velocity="(0.2,2.0,5)"` is provided, `FabSIM` interprets this as `{"particle_initial_velocity": "\"(0.2"}` and discards the rest of the tuple. I couldn't find any obvious way around this, so I'm suggesting that for the command line definition of our sweep bounds the list be given with a `/` separating the parameters. This is then interpreted into the correct form inside the `neso_ensemble` task. The above example would then be: `particle_initial_velocity=0.2/2.0/5`. If that sounds acceptable, I can update the documentation to reflect this.

That's a pain with regards to FabSim's argument handling, but your proposed solution seems reasonable. The only thing I would suggest is to use the colon `:` as the separator character as this is already used in Python slices for a similar purpose - for example `1:10:2` represents the slice containing indices `(1, 3, 5, 7, 9)` - so might be a bit more intuitive for users to remember, and also `/` possibly suggests either division or a date separator (for floating point end-values seems unlikely to be interpreted this way, but for example my first though on seeing something like `1/10/19` would be 1st October 2019 not `(1., 1.5, 2., ..., 9.5, 10.)` 😅).

FabNESO/ensemble_tools.py
```diff
@@ -78,33 +78,36 @@ def _product_dict(input_dict: dict) -> Iterator[dict]:
...
+            for i in range(n_dir_parameter)
+        ]
+        for key, (low, high, n_dir_parameter) in parameter_dict.items()
     }
     # Compute Cartesian products of all parameter value combinations plus grid indices
+    [n_dir[2] for n_dir in parameter_dict.values()]
     for parameter_values, indices in zip(
         _product_dict(parameter_grids),
-        indices_iterator(n_dirs, len(parameter_grids)),
+        indices_iterator([n_dirs[2] for n_dirs in parameter_dict.values()]),
```
matt-graham: Not important here, but in case it's helpful in the future, you can avoid the overhead of creating a list here in the list comprehension by instead using a [generator expression](https://peps.python.org/pep-0289/) as the `indices_iterator` method just needs an object it can iterate over not necessarily a list or other sequence (would also need to update the type-hint to `Iterable` or similar to satisfy `mypy` I think). In this case the performance difference will be negligible as the number of items will be small, but it can be a useful approach to avoid creating intermediate lists when working with larger structures.

FabNESO/ensemble_tools.py
```diff
@@ -78,33 +78,36 @@ def _product_dict(input_dict: dict) -> Iterator[dict]:
...
-        key: [calculate_parameter_value(n_dirs, low, high, i) for i in range(n_dirs)]
-        for key, (low, high) in parameter_dict.items()
+        key: [
+            calculate_parameter_value(n_dir_parameter, low, high, i)
+            for i in range(n_dir_parameter)
+        ]
+        for key, (low, high, n_dir_parameter) in parameter_dict.items()
     }
     # Compute Cartesian products of all parameter value combinations plus grid indices
+    [n_dir[2] for n_dir in parameter_dict.values()]
```
matt-graham:
```suggestion
```
This looks like it might be a leftover from some earlier iteration as not assigned to anything?

FabNESO/ensemble_tools.py
matt-graham:
```suggestion
        indices_iterator([n_dir for *_, n_dir in parameter_dict.values()]),
```
Might be a bit more explicit that we just want to ignore the first values in each list in `parameter_dict.values()` and extract the final `n_dir` value (rather than naming whole list `n_dirs`). `_` is commonly used as a dummy variable name to assign unused values in Python and the `*` operator here will unpack multiple arguments in to this variable.

FabNESO/make_sweep_dir.py
```diff
@@ -4,7 +4,7 @@
 from ast import literal_eval
 from pathlib import Path
 
-from .ensemble_tools import create_dict_sweep, create_dir_tree
+from FabNESO.ensemble_tools import create_dict_sweep, create_dir_tree
```
matt-graham: Did the relative import not work in some context here?

FabNESO/make_sweep_dir.py
```diff
@@ -70,10 +70,14 @@ def main() -> None:
         if not isinstance(parameter_dict, dict):
             msg = "Did not receive a dict as input for parameter_dict"
             raise ValueError(msg)
+        # If we define n_dirs in the command line argument, override any
+        # dict parameters we didn't assign n_dirs to
+        for _parameter, values in parameter_dict.items():
```
matt-graham:
```suggestion
        for _, values in parameter_dict.items():
```
`_` is conventionally used as dummy variable name for unused variable assignments

FabNESO/tasks.py
```diff
@@ -90,16 +91,47 @@ def neso_ensemble(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
+        parameter_scans is a set of parameters to sweep over. A slash separated list of
+        lower bound, upper bound, and steps.
```
```suggestion
        **parameter_scans: The set of parameters to sweep over. A slash separated list of
            lower bound, upper bound, and steps.
```
To conform with [Google style docstring convention](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) (will make sure things render nicely in HTML documentation once we address #16).

FabNESO/tasks.py
```diff
@@ -90,16 +91,47 @@ def neso_ensemble(
...
-    fab.run_ensemble(config, sweep_dir)
+    with temporary_context as temporary_config_directory:
+        if parameter_scans != {}:
+            temporary_config_path = Path(temporary_config_directory)
+            # Because FabSIM is a bit weird with commas, build the dict here
+            parameter_scan_dict = {
+                parameter: (
+                    float(values.split("/")[0]),
+                    float(values.split("/")[1]),
+                    int(values.split("/")[2]),
```
matt-graham: I would factor out the parsing of the strings here to a separate function to avoid having to split multiple times, something like
```Python
def _parse_parameter_scan_string(
    parameter_scan_string: str, delimiter: str,
) -> tuple[float, float, int]:
    start, end, n_steps = parameter_scan_string.split(delimiter)
    return float(start), float(end), int(n_steps)
```

This should already raise a relatively informative error if the string is not of the expected format (as won't be able to unpack split string into three arguments) but could also explicitly check length of split string and raise an error if not equal to 3.

pyproject.toml
```diff
@@ -19,6 +19,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
+    "numpy",
```
matt-graham: If `numpy` is only a test time dependency it should be in `dev` dependencies.

tests/test_ensemble_tools.py
```diff
@@ -4,6 +4,7 @@
 from pathlib import Path
 from typing import TypedDict
 
+import numpy as np
```
matt-graham: To avoid adding `numpy` as a dependency as we only use `prod` function, we could instead import
```Python
from functools import reduce
from operator import mul
```
and then use `reduce(mul, list_of_numbers_to_multiply)` to compute products.

tests/test_ensemble_tools.py
```diff
@@ -248,7 +258,7 @@ def test_create_dict_sweep(
             parameter = list(parameter_dict.keys())[i]
             scan_range = parameter_dict[parameter]
             para_value = calculate_parameter_value(
-                n_dirs, scan_range[0], scan_range[1], indices[i]
+                scan_range[2], scan_range[0], scan_range[1], indices[i]
             )
```
matt-graham: Unpacking `parameter_dict[parameter]`  with something like `start, end, n_dir = parameter_dict[parameter]` would give more meaningful names to different entries and make intention of subsequent `calculate_parameter_value` call clearer.

matt-graham (changes requested):

FabNESO/make_sweep_dir.py
dleggat: It wasn't working when I was running the script interactively, but that that shouldn't really be an issue, so I've put it back to being a relative import.

FabNESO/make_sweep_dir.py
dleggat: With this one in particular I've just changed it to:
```python
for values in parameter_dict.values():
``` 
but your remark about the dummy variables is noted and taken onboard.

FabNESO/tasks.py
```diff
@@ -91,8 +99,8 @@ def neso_ensemble(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
-        parameter_scans is a set of parameters to sweep over. A slash separated list of
-        lower bound, upper bound, and steps.
+        **parameter_scans: The set of parameters to sweep over. A slash separated list
```
```suggestion
        **parameter_scans: The set of parameters to sweep over. A colon separated list
```

FabNESO/tasks.py
```diff
@@ -105,11 +113,7 @@ def neso_ensemble(
             temporary_config_path = Path(temporary_config_directory)
             # Because FabSIM is a bit weird with commas, build the dict here
             parameter_scan_dict = {
-                parameter: (
-                    float(values.split("/")[0]),
-                    float(values.split("/")[1]),
-                    int(values.split("/")[2]),
-                )
+                parameter: (_parse_parameter_scan_string(values, ":"))
```
```suggestion
                parameter: _parse_parameter_scan_string(values, ":")
```
Don't think the parenthesis here is doing anything

matt-graham (changes requested):
Looks good other than two very minor points.

matt-graham (approved):
All good to merge!

PR #32, opened 2023-12-18
Update conditions files to match original NESO examples

matt-graham:
Updates the example `conditions.xml` files in `config_files` to exactly (bar tab to space subsititutions) match the `two_stream` and `2Din3D-hw` examples in https://github.com/ExCALIBUR-NEPTUNE/NESO/blob/main/examples/Electrostatic2D3V/two_stream/two_stream_conditions.xml and https://github.com/ExCALIBUR-NEPTUNE/NESO/blob/main/examples/H3LAPD/2Din3D-hw/hw.xml respectively.

The current files reduced the number of simulated time steps, presumably to allow for quicker iteration when testing, and for the `2Din3D-hw` case was based on the simpler `2Din3D-hw_fluid_only` variant. As we now support overriding the parameters when running the task, I think having the conditions match the original NESO examples for consistency makes more sense, with it still possible to quickly override the parameters when running to reduce the compute time required.

### DIFF
--- a/config_files/2Din3D-hw/conditions.xml
+++ b/config_files/2Din3D-hw/conditions.xml
@@ -7,7 +7,7 @@
         The composite index for the domain is expected to be 0 if the mesh was generated from the included .geo file
      -->
     <EXPANSIONS>
-       <E COMPOSITE="C[0]" NUMMODES="7" TYPE="MODIFIED" FIELDS="ne,w,phi" />
+       <E COMPOSITE="C[0]" NUMMODES="6" TYPE="MODIFIED" FIELDS="ne,w,phi,ne_src" />
     </EXPANSIONS>
 
     <CONDITIONS>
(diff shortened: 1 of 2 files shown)


dleggat (approved):
All looks good to me. I assume the other parameter that this is missing from the `2Din3D-hw/conditions.xml` file;

```
            <!-- Turn on energy, enstrophy output -->
            <P> growth_rates_recording_step = 1 </P>
```
is only there for writing output, so can just be added from the command line if we need this information?

matt-graham:
> All looks good to me. I assume the other parameter that this is missing from the `2Din3D-hw/conditions.xml` file;
> 
> ```
>             
>             <P> growth_rates_recording_step = 1 </P>
> ```
> 
> is only there for writing output, so can just be added from the command line if we need this information?

Ah good spot, looking at [the blame](https://github.com/ExCALIBUR-NEPTUNE/NESO/blame/main/examples/H3LAPD/2Din3D-hw/hw.xml) for the original source file in the NESO repository it looks like these lines were added in a more recent commit so it looks like I didn't pull in the latest changes on NESO before copying over. I'll update this now, though I guess it's likely they will still end up diverging at some point.

PR #33, opened 2023-12-18
Use `run_command` in template script

matt-graham:
Updates the `templates/neso` template script to prefix the solver executable run command with the value of the `run_command` variable. This allows the user to configure a machine specific value of `run_command` to use to dispatch the solver runs on different machines. By default [on `localhost` FabSim3 defaults `run_command` to `mpirun -np $cores`](https://github.com/djgroen/FabSim3/blob/62ca215febb2ef5112a7f5a59574ae3db465a1e7/fabsim/deploy/machines.yml#L64). Here we also the default value (in the example configuration script) for `run_command` on `localhost` to ~~`"OMP_NUM_THREADS=1 mpirun -n $cores -map-by core -bind-to hwthread"` in the `machines_FabNESO_user_example.yml`, which will ensure no OpenMP multithreading is used and to  `"srun --distribution=block:block --hint=nomultithread"` on `archer2`~~ `"mpirun -n $cores -map-by core -bind-to hwthread"` in the `machines_FabNESO_user_example.yml` and export `OMP_NUM_THREADS=1` in the template script to disable threading to avoid over-subscription of CPU cores when parallelising with MPI. 

Edit: Default `archer2` run command now updated in FabSim so specification here removed.

### DIFF
--- a/machines_FabNESO_user_example.yml
+++ b/machines_FabNESO_user_example.yml
@@ -10,4 +10,4 @@ default:
 
 localhost:
   neso_bin_dir: "change-me" # path to directory containing built NESO solver executables
-  run_command: "OMP_NUM_THREADS=1 mpirun -n $cores -map-by core -bind-to hwthread"
+  run_command: "mpirun -n $cores -map-by core -bind-to hwthread"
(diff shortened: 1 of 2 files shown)


dleggat (approved):
Looks good! This needed a bit of a cleanup, and the prefix `run_command` should help running remotely.

PR #34, opened 2023-12-19
Expose task arguments to set job resources

matt-graham:
Adds arguments `processes`, `nodes`, `cpus_per_process` and `wall_time` to both `neso` and `neso_ensemble` tasks to allow setting respectively number of processes to run job on, number of nodes to run processes over (on multi-node systems), number of processing units per process and maximum wall clock time (on systems with a job scheduler). Some basic checking that passed arguments are sensible is performed and they are translated to corresponding FabSim environment variable names which are used to set options in run command / scheduler job submission script.

### DIFF
--- a/FabNESO/tasks.py
+++ b/FabNESO/tasks.py
@@ -25,7 +25,7 @@ def _try_convert_to_int_and_check_positive(value: str | int, name: str) -> int:
     try:
         value = int(value)
     except ValueError as e:
-        msg = f"{name} is not a valid an integer literal: {value}"
+        msg = f"{name} is not a valid integer literal: {value}"
         raise ValueError(msg) from e
     if value <= 0:
         msg = f"{name} must be a positive integer: {value}"
@@ -185,7 +185,7 @@ def neso_ensemble(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
-        processes: Number of processes to run each job in ensemble on.
+        processes: Number of processes to run in each job in the ensemble.
         nodes: Number of nodes to run each job in ensemble on. Only applicable when
             running on a multi-node system.
         cpus_per_process: Number of processing units to use per process for each job in
--- a/FabNESO/tasks.py
+++ b/FabNESO/tasks.py
@@ -4,6 +4,7 @@ Task definitions for FabNESO plug-in to FabSIM software toolkit.
 Defines tasks for running simulations using Neptune Exploratory Software (NESO).
 """
 
+import re
 import shutil
 from contextlib import nullcontext
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dleggat (approved):
I was a bit confused by the 'cores' and 'processes' naming conventions that didn't seem to match up, but then I found the comment about FabSim calling MPI processes 'cores', and it began to make more sense. Seems weird, but I suppose it makes sense to hide that from the users at our end.

Otherwise everything works fine for me barring some small typos, so I'm happy with it.

FabNESO/tasks.py
```diff
@@ -20,13 +21,81 @@
 fab.add_local_paths("FabNESO")
 
 
+def _try_convert_to_int_and_check_positive(value: str | int, name: str) -> int:
+    try:
+        value = int(value)
+    except ValueError as e:
+        msg = f"{name} is not a valid an integer literal: {value}"
```
dleggat:
```suggestion
        msg = f"{name} is not a valid integer literal: {value}"
```

FabNESO/tasks.py
```diff
@@ -37,10 +106,20 @@ def neso(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
+        processes: Number of processes to run on.
```
dleggat: Sounds like this was written for 'cores' and then replaced with processes - would `Number of processes to run.` sound more accurate?

FabNESO/tasks.py
```diff
@@ -99,9 +185,19 @@ def neso_ensemble(
         solver: Which NESO solver to use.
         conditions_file_name: Name of conditions XML file in configuration directory.
         mesh_file_name: Name of mesh XML in configuration directory.
+        processes: Number of processes to run each job in ensemble on.
```
dleggat: Similarly this sounds a bit weird.
```suggestion
        processes: Number of processes to run in each job in the ensemble.
```

FabNESO/tasks.py
matt-graham: Agreed your suggestion is better, will update!

matt-graham:
> I was a bit confused by the 'cores' and 'processes' naming conventions that didn't seem to match up, but then I found the comment about FabSim calling MPI processes 'cores', and it began to make more sense. Seems weird, but I suppose it makes sense to hide that from the users at our end.

Sorry just remembered I forgot to reply to this! Agree that the inconsistency between usages of cores and processes is not ideal and as you say was largely trying to hide the details of how things are defined in FabSIM from the user, but still wasn't entirely satisfied with the solution I came up with. There is some downside in that other FabSIM plugins I think directly expose these additional arguments by allowing passing in arbitrary keyword arguments to the task and so would use a different argument to set the number of processes, but particularly for the context of running on HPC systems I thought using cores when instead processes is meant could end up being misleading to users. I also briefly considered `tasks` instead of processes which follows the terminology used in Slurm a bit better but I thought processes was probably the more widely recognised term.

PR #36, opened 2024-01-05
Add autogenerated HTML API documentation built with Sphinx

### LINKED ISSUE #16: Create HTML documentation on GitHub pages

matt-graham:
Fixes #16 

Sets up infrastructure to allow building package API documentation for FabNESO using Sphinx and building HTML output as a part of a GitHub Actions workflow.

As `sphinx-autodoc` needs to be able to import modules being documented, I've added in a slightly hacky solution to allow importing the `FabNESO.tasks` module by creating a placeholder `fab` namespace if FabSim3 is not installed (with it still not possible to automate installation of FabSim3 with `pip` atm). I've added a check function to the beginning of the tasks which checks FabSim is actually installed / importable before trying to run and gives an informative error message if not.


dleggat:
Is there a way I can preview the generated docs? I can see from the checks section that they generate correctly, but the directory they're put into is, I guess, just one created during the workflow there.

matt-graham:
If you checkout this branch locally and run `tox -e docs` that should perform the equivalent steps for building the docs as in the GitHub Actions workflow (with the workflow just additionally dealing with checking out the repository, setting up dependencies and pushing HTML output to a `gh-pages` branch if triggered by a push to `main`).

dleggat:
Ah, I'd gotten as far as running tox on the branch, but missed the extra argument step. Cool, all looks good then!

dleggat (approved):

PR #41, opened 2024-01-24
Moving dependencies to requirements.txt

dleggat:
Moving package dependencies to a `requirements.txt` to fix #37

PR #42, opened 2024-01-24
Restructure neso_vbmc input parameters

### LINKED ISSUE #38: Allow passing in `observation_noise_std` as argument to `neso_vbmc` task
Passing in `observation_noise_std` in as an argument to the task would be the easiest way to let user change.

_Originally posted by @matt-graham in https://github.com/UCL/FabNESO/pull/30#discussion_r1453759699_

### LINKED ISSUE #39: Make `reference_field_file` a required argument to `neso_vbmc` task
Rather than having `reference_field_file` default to an empty string, I'd say probably better to not give a default and have this as a required argument as that might suggest it doesn't need to be set.

_Originally posted by @matt-graham in https://github.com/UCL/FabNESO/pull/30#discussion_r1453757850_

dleggat:
Exposed noise value to the command line to resolve #38 

Moved reference_field_file to be a mandatory input to resolve #39 

I took the liberty of ignoring lint's complaints about there now being too many input parameters, as we are expecting to be reading these in from the command line prior to execution.


pyproject.toml
matt-graham, line 1: I think you've also included the changes from #41 here - probably doesn't matter as long as that is merged in first.

FabNESO/tasks.py
```diff
@@ -309,18 +314,22 @@ def neso_vbmc(
 
     Args:
         config: Directory with ensemble configuration information.
+        reference_field_file: Name of a numpy file that holds a reference field
+            measurement for the calibration run.
```
```suggestion
        reference_field_file: Name of a NumPy .txt file that holds a reference
            field measurement for the calibration run.
```
Making the expected file type explicit probably helpful as NumPy can output to a few different formats (`.npy` / `.npz` / `.txt`).

matt-graham (changes requested):
One small suggestion otherwise looks good thanks!

matt-graham (approved):

